PR #3, opened 2025-03-21
Company Scrip

taydeo:
you can now bribe easier

### DIFF
--- a/src/design-proposals/scrip.md
+++ b/src/design-proposals/scrip.md
@@ -1,0 +1,96 @@
+# Company Scrip
+
+| Designers | Implemented | GitHub Links |
+|---|---|---|
+| taydeo | :x: No | TBD |
+
+## Overview
+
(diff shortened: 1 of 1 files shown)


misterghast:
I think, to replace the ACC wire snipping to use the vending machine without having the proper ID access, pulsing it with a multitool should allow buying items from vending machines from anyone (with the required funds) regardless of ID job accesses, but only for 30 seconds so it's in line with other multitool pulsing effects. I like the idea of snipping disabling it outright, though. It doesn't make sense that the vending machines would just allow anyone to use it if the entire authentication system was completely disabled.

mkanke-real:
Will your balance be carried between rounds so someone can make a long term goal of a character just getting "rich" on company script? 
-If so would the balance carrier between characters?

Before round start will script be used to buy trinkets or items so you can go into a round with less script but more goodies or vis versa?

taydeo:
> Will your balance be carried between rounds so someone can make a long term goal of a character just getting "rich" on company script?

I'm not too sure if this is something we want to do.

PR #5, opened 2025-03-24
how2contribute4dummies

corresp0nd:
hopefully there's no typos cuz im kinda eepyyyy.... also feel free to nitpick explanations or suggest stuff i should add; i would prefer if this is as easy to understand as possible


taydeo:
You should also mention the Python issue with Windows. Fuck Microsoft.

PR #10, opened 2025-06-13
corporate secrets

corresp0nd:
tumblr_4c25c02dcd03566551a3ac23fd541981_414b5968_400

### DIFF
--- a/src/design-proposals/command-sop.md
+++ b/src/design-proposals/command-sop.md
@@ -1,0 +1,10 @@
+# Command SOP
+
+1. Command members are to ensure the safety of the entire station, each primarily responsible for their own Department and respective duties;
+2. Command members are to ensure the wellbeing and dignity of Nanotrasen, any station events should be resolved to the best of their ability; 
+3. Command members may not slander or libel Nanotrasen;
+4. Command members must ensure that Nanotrasen corporate secrets remain confidential, and may reprimand or punish fellow command members who threaten this confidentiality; 
+5. In the event of a confirmed threat to the station or Nanotrasen as a whole, command members may temporarily reveal related corporate secrets to crewmembers; 
+6. Command members must report such confidentiality breaks as soon as the event is resolved; 
+7. Command members must follow any Central Command officer's (such as the Nanotrasen Representative, Internal Affairs Agent, or the Magistrate) direction on corporate secrets; 
+8. Command members must accept any fine, punishment, demotion, or death when a Central Command officer informs them of a wrongful break of their confidentiality agreement. 
--- a/src/design-proposals/command-sop.md
+++ b/src/design-proposals/command-sop.md
@@ -1,7 +1,7 @@
 # Command SOP
 
 1. Command members are to ensure the safety of the entire station, each primarily responsible for their own Department and respective duties;
-2. Command members are to ensure the wellbeing and dignity of Nanotrasen, any station events should be resolved to the best of their ability; 
+2. Command members are to ensure the wellbeing and dignity of Nanotrasen and its related assets; 
 3. Command members may not slander or libel Nanotrasen;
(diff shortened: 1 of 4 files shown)


wolfy9011:
I feel there should be other ways to reveal the metashield besides just command telling you, jobs like security should know about certain antags protected by said shield, including normal traitors and theifs.

What is a crewmember that is under this new metashield supposed to do when they get walked up on by a thief or somesuch saying they work for the syndicate and showing them illegal stuff that normal crew CANT obtain. (Hell even using the uplink infront of someone would do that as the item would just appear in their hand suddenly)

Just to TLDR: There should be more ways to reveal the shield besides just being told about it by command members

corresp0nd:
re: wolfy9011

> I feel there should be other ways to reveal the metashield besides just command telling you, jobs like security should know about certain antags protected by said shield, including normal traitors and theifs.

i agree security should know more, it's one of the things i want to touch up on. originally, i didn't include them having any sort of lifted metashield because i really wanted to focus on the aspect of "jail people for what they did, not what antag they are." with the removal of EOC, though, i don't think such a hard rule is necessary for security. 
i'm thinking of making a "training manual," an in game book for security, that would tell them what they're allowed to know without breaking the shield. ideally, this would tell security that there are potential threats (ex. "the syndicate is a highly dangerous organization"), but flavor that with NT claiming that nothing ever happens ("but since our station is in the middle of nowhere, we probably won't ever see them!"). something to give leeway to security so they can do their job, but still reinforce the expectations we have for roleplay regarding antags. 

> What is a crewmember that is under this new metashield supposed to do when they get walked up on by a thief or somesuch saying they work for the syndicate and showing them illegal stuff that normal crew CANT obtain. (Hell even using the uplink infront of someone would do that as the item would just appear in their hand suddenly)

for the uplink bit, that would be entirely metashielded. while you could know that they have an uplink ooc, ic you should assume they just took it out of their bag. this rewrite emphasizes that average crewmembers don't know *anything* about antag objectives, tools, etc. 
however, if you do find yourself in the scenario that someone is undeniably an antag, the ideal scenario is to tell security/command about it. you could even pressure them into revealing the metashield by showing the "proof" that you have seen. basically just roleplay that you don't know what's happening, but try to force security/command into responding. or don't tell them at all, and let yourself be an accomplice. it is sec/command's job to keep the station safe, not yours as a crewmate. by making the metashield more strict like this, i'm hoping to encourage more roleplay and less of a focus on winning/losing. 



re: ferynn 
> Is there ever a point where breaking the metashield on “minor” stuff is allowed? I was just thinking about mysterious fibers, and chameleon gear and the like. It’s generally unlikely that such things will ever be used in a way to threaten serious crew harm, so, barring edge cases, would command ever tell the detective what those mean?

yes, command would be able to tell relevant people if they so deem fit. in this particular case, mysterious fibers/chameleon gear can mean a lot of things. if it's a thief wearing chameleon gloves, then it isn't a big issue. det would scan a stolen item, and bring the report to hos to ask what that's all about. hos can either say "nothing ever happens" or lift the metashield and tell the det that some thieves may have illegal gear that they should investigate (or if they think it's something more major, like a ninja, stealth nukies, etc, then lifting the metashield is also okay). either way, the way the det responds shouldn't change very much, they should still investigate the primary issue (something being stolen or hacked into) rather than trying to hunt based on antag status. im going to try and refine it more to reflect this idea; command can reveal the shield at any time as long as they deem it fit. 



and to try and reemphasize, because i don't think i explictly stated this: the goal of this change is to change antag-interaction from being focused on finding the traitor (think gameplay loops of other traitor games like town of salem, among us, mafia) to focusing on traitors creating more drama in the round (i dunno how i could explain this. ace attorney gameplay?). the shift shouldn't end when the traitor is "found." traitors should be able to continue doing evil deeds even if their initial plan falls through, and security/crew should be focused on catching crime rather than antagonists.

ferynn:
Also I realized I forgot to add, I do like what you’re doing and working towards here. I think it has a lot of potential. Was just thinking about edge cases

PR #11, opened 2025-06-30
Lasers / Armory Overhaul 

DevilishMilk:
If there's any interest here, I'd be happy to discuss in detail. Main cooperation required would be on the mapping sides, but this is all fairly simple to actually change. The specifics will need work.

### DIFF
--- a/src/design-proposals/lasers.md
+++ b/src/design-proposals/lasers.md
@@ -32,19 +32,31 @@ Ranged weapons, due to their inherent safety and power over using melee, should 
 
 ## Additions
 
-### Temperature Gun
+### Map into Armory
+
+#### Temperature Gun
 See https://github.com/space-wizards/space-station-14/pull/35447 for implementation. Main difference: this will be an Armory weapon and not researchable.
 
-### Ion Rifle
+#### Ion Rifle
 A powerful three-shot rifle that deals 50 ion damage on a direct hit as well as a 3x3 tile weak EMP where it lands. Very powerful against cybernetic threats, such as borgs, IPCs, or mechs, but only tactically useful against other energy-weapon users.
 
-### Energy Gun
+#### Energy Gun
 A battery-powered carbine (can fit in the backpack) which can fire either disabling or laser bolts, with less charge than the base Laser Rifle. It has 15 charges.
 
-### Miniature Energy Gun
+#### Miniature Energy Gun
 The replacement for both the Officer's disabler and pistol, this is simply a smaller version of the Energy Gun. It can fit in your pocket and has 12 charges, which can be used to disable or laser.
 
 Cadets will also start with this, but it will be locked on Disable until the childproof plastic lock is broken using a screwdriver.
+
+### Possible Additions (to be added to Research, primarily)
+
+#### Blue Laser Rifle
+Same as the stock laser rifle but it deals cold damage because it's blue.
+
+#### Force-bolt Pitcher
+An energy weapon that fires slow-moving but long-lasting spheres of energy. This deals blunt and some stamina damage.
+
+####
 
 ## Changes to current Features / Gamestate
 
@@ -59,6 +71,8 @@ The addition of a disabler safe to all armories for emergency Command-arming rea
 The removal of the WT-550 from the Head of Security office.
 
 The removal of disablers from Sec officer lockers.
+
+The removal of ballistic ammo that no longer corresponds to armory/on-station weaponry. Keep rifle, revolver, and shotgun ammo printing for the Chaplain, Detective, and armory respectively.
 
 Overall, Armory would now contain:
 - 1 disabler safe (containing 5 disablers)
--- a/src/design-proposals/lasers.md
+++ b/src/design-proposals/lasers.md
@@ -1,4 +1,4 @@
-# Company Scrip
+# Lasers and Armory Overhaul
 
 | Designers | Implemented | GitHub Links |
 |---|---|---|
--- a/src/design-proposals/lasers.md
+++ b/src/design-proposals/lasers.md
@@ -56,7 +56,8 @@ Same as the stock laser rifle but it deals cold damage because it's blue.
 #### Force-bolt Pitcher
 An energy weapon that fires slow-moving but long-lasting spheres of energy. This deals blunt and some stamina damage.
 
-####
+#### Inducer
+A portable recharger. It takes battery-cells and charges at a very slow rate. Possibly upgradable with Science.
 
 ## Changes to current Features / Gamestate
 
--- a/src/design-proposals/lasers.md
+++ b/src/design-proposals/lasers.md
@@ -115,40 +115,4 @@ This should ideally encourage players in combat to fight in more up-and-personal
 
 This is all possible to do with purely .yml and mapping changes. Coordination will be required from mapping maintainers for armory changes.
 
-More input will be required from the maintainers, particularly concerning the specifics of numbers, and the specifics of the Rollout following this section.
-
-## Rollout
-
-### Week 1
-
-- Laser projectile changes, damage tweaks to all current laser weaponry.
-- Minor (-10%) nerf to heat resist on all nukie suits.
-- Initial rollout of Miniature Energy Gun.
-- Addition of the Miniature Energy Gun to Cargo.
-- Miniature Energy Gun will replace Sec Officer's pistol and disabler, as well as the HoP's disabler plus.
-- Initial rollout of the Energy Gun.
-- The addition of Energy Gun crates to Cargo.
-- Replacing the Enforcer crate in Cargo with a Kammerer crate.
-- Removal of the SMG crate in Cargo.
-- Removal of the pistols crate in Cargo.
-
-### Week 2
-
-- Additional laser and armor tweaks as needed.
-- Initial rollout of the Temperature Gun.
-- The addition of the Temperature Gun to the armory.
-- The removal of ion damage from the disabler.
-- Initial rollout of the Ion Rifle.
-- The addition of the Ion Rifle to the Armory.
-- The removal of self-recharging from non-Captain's guns.
-
-### Week 3+ 
-
-- Additional laser and armor tweaks as needed.
-- Removal of the WT-550 from the HoS' office.
-- The replacement of the Lecter and Drozd safes with the Energy Gun & disabler safes in the Armory.
-- The removal of self-recharging from non-Captain's guns.
-
-### Week 4+ 
-
-- Monitor and make additional balancing tweaks as necessary.
+More input will be required from the maintainers, particularly concerning the specifics of numbers.
--- a/src/design-proposals/lasers.md
+++ b/src/design-proposals/lasers.md
@@ -1,0 +1,154 @@
+# Company Scrip
+
+| Designers | Implemented | GitHub Links |
+|---|---|---|
+| Miiish | :x: No | TBD |
+
+## Overview
+
+This design doc outlines the removal of current ballistic weapons in the armory in favor of safer, more retro, family-friendly weaponry designed to melt station-ending threats.
+
+## Background
+
+Ballistic weaponry from an in-universe common-sense/immersion perspective is not ideal whatsoever when bullets can break windows and space (very dangerous!!!) sections of the station, potentially killing crew if no protections are in place.
+
+This design doc is made from the perspective of both allowing melee weapons to be more viable and to unify NanoTrasen standard weaponry in a way that makes more sense.
+
+They are additionally the primary killing tools of the Syndicate (bad guys!!), and this works to differentiate NanoTrasen brand killing tools (good) and Syndicate brand evil killing tools (bad), as well as allow unique balancing around factional weaponry. Found bullet casings in maintenance? You probably can't blame that on a trigger-happy cadet anymore.
+
+## Why this is Good for the Game
+
+These changes all aim to serve a few purposes:
+
+Aesthetics: Energy weapons are cool and retro sci-fi-y, and very distinguishable as opposed to having 5 generic Gun sprites.
+
+Melee, especially unique antagonist melee such as the energy sword and armblade, should be more viable as a method of killing, as a high risk-reward profile. The overall increase of Time-To-Kill on ranged weaponry should aim to incentivize the use of more 'up close and personal' killings and encourage antagonists to take on Security more.
+
+The splitting up of Sec lethals should also encourage Security to use their other tools, like truncheons, more for better damage profiles instead of using their gun as a metaphorical hammer for everything.
+
+This also aims to bring Sec more 'back to Earth', so to speak. By lowering the amount of pure weaponry they have access to roundstart, we curb the constant Sec - Antag cat-and-mouse balancing game that leads to intense powercreep, and makes crew cooperation and interdepartmental play somewhat more useful in ending station-wide threats, such as the Blob. In short, this helps allow for antagonist abilities to be imposing without having to be overwhelmingly powerful or durable.
+
+Ranged weapons, due to their inherent safety and power over using melee, should require a bit more tactical use and forethought than just pointing in a direction and killing whatever's there. Lasers allow for more of this planning through the use of, for example, constructing windows to prepare for a threat. The lowered amount of charges and damage-per-charge/magazine also requires more forethought when bringing them out. Why use a gun when I can relatively safely use melee here?
+
+## Additions
+
+### Temperature Gun
+See https://github.com/space-wizards/space-station-14/pull/35447 for implementation. Main difference: this will be an Armory weapon and not researchable.
+
+### Ion Rifle
+A powerful three-shot rifle that deals 50 ion damage on a direct hit as well as a 3x3 tile weak EMP where it lands. Very powerful against cybernetic threats, such as borgs, IPCs, or mechs, but only tactically useful against other energy-weapon users.
+
+### Energy Gun
+A battery-powered carbine (can fit in the backpack) which can fire either disabling or laser bolts, with less charge than the base Laser Rifle. It has 15 charges.
+
+### Miniature Energy Gun
+The replacement for both the Officer's disabler and pistol, this is simply a smaller version of the Energy Gun. It can fit in your pocket and has 12 charges, which can be used to disable or laser.
+
+Cadets will also start with this, but it will be locked on Disable until the childproof plastic lock is broken using a screwdriver.
+
+## Changes to current Features / Gamestate
+
+### Armory (Mapping)
+
+The removal of all ballistic weaponry from the Armory, in favor of the new laser-based weaponry.
+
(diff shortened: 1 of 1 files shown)


DevilishMilk:
Thank you for the thought-out response (curse of copy+pasting the template for formatting FUCK it's fixed), and yeah these changes all have heavy balance implications that I think are necessary and healthy for the game, Security, Antagonists, and Crew in the long run but yes, they will take some time to get right.

I think this is a good point to pivot from the current Wizard's Den / Goob philosophy of 'is this gun / antagonist / thing underperforming? Let's just buff the everloving hell out of it until people stop complaining' - I think that's unhealthy and is the textbook case of power creep as can be seen here with the current state of the Armory. There's no current incentive to really research the Arsenal except for a select few weapons, since the Armory's already got quite literally everything you need for every threat.

While the intention here is not to nerf either Sec or the Crew's overall ability to respond to antagonist threats, the antagonists in question that are typically responded to with lethals (think Nukies, Dragons, Blobs primarily) will absolutely need some tweaking. I laid out some ideas in the proposal relating to them, whether that's adjusting health or resistances, but it's a nonexhaustive list and I think it can only really be tweaked when we see this in action.

Your point with the magazines both here and in Discord have me convinced that some researchable laser magazines may be interesting as a research item.

The intent of this is to have antagonists powerful in their own rights without having to balance around how fast they'll go down to bullets, and to give Sec a better, healthier roundstart foundation on which we can balance Arsenal research, nonlethal tools, and Cargo ordering.

GreyMario:
I still don't like the idea of completely removing ballistics from Nanotrasen gear.

Also, switching purely to heat based weaponry will just result in funny things like nukie medics doing nothing but preparing anti-burn meds, on-site antags that take a few hits from a seccie having no way to heal their burns (they would be able to gauze up bullet wounds with scraps of cloth), and Atmos traitors being completely invincible.

Teasq:
"Removal of the SMG crate in Cargo" - self-explanatory," 
The WT-550 is a majorly overtuned weapon in my opinion, so I think it's removal from standard play is a good thing overall.

I do think both removing the self-recharging capability of the multiphase and the Wt-550 will have a major impact on security's ability to respond to threats though, and they should probably be spaced out to see how the removal of the WT-550 goes first. For the multiphase, maybe instead of complete removal just make it noticeably slower than the cap's recharging? It'll still have its unique edge because of its multifunctionality, but won't be plainly better.

packmore:
i actually quite like this idea on paper. i really quite enjoy melee weapons being given more love here. but there's a few specifics i think should be hashed out before i'm fully on board with this doc.

- shotguns are arguably the most dangerous guns when it comes to breaking windows. short bursts of bullets means there's very little time to correct yourself if you accidentally fire at a window. unlike most automatic guns, where you can just stop aiming at the window and correct yourself, shotguns are far more likely to bust them open. it makes very little sense that NT would remove automatic ballistics and not shotguns, and to be honest, i would rather see more energy shotguns than kammerers. 
  - ballistics don't have to be _entirely_ removed, but if the point is to remove the risk associated with ballistics, then the armory simply cannot come stocked with any, or else it stops making sense. having low-power ones be orderable from cargo, or researched by science, makes the most sense to me.
  - the security shell gun could unironically have a use if you remove ballistic shotguns, as well as mostly being accurate in-lore. 
  - energy shotguns already exist. you could very easily make the warden's specialized to have a higher capacity, or have it be the only shotgun that can use disabling projectiles, or something similar.
  - shotguns are already the most oppressive ballistic weapon in the game. keeping them intact would basically just invalidate the rework, especially because security is likely to immediately order enough crates of them to equip the whole security team in the event of something dangerous coming up. 
- things like the energy shield would need to be fully removed. the DEsword's huge reflect chance would likely need to be nerfed as well. when you're firing fewer shots that are easier to dodge and do less damage, you can't really afford to have such huge ways of total projectile nullification. 
- you could very easily argue that the reason NT uses ballistics currently is due to the cost of production. energy weapons require more expensive materials to create, as well as just being much more complex than ballistics. i don't know if this actually affects anything, but it's worth pointing out.
- melees may require small buffs across the board, if you're going to more heavily emphasize their usage. not sure on this point, though. would require testing.
- disablers probably already need a nerf, but doubly so here. if the idea is to push people to use melee more, having a weapon that completely invalidates the existence of the stun baton is really counter-productive.
- the arsenal tech tree would need to be reworked to some degree for this to work. as-is, science has very little reason to ever research arsenal tech, and if this doc was implemented fully, they would have even less of a reason to do so. the laser cannon and gygax are the only two techs that security would ever seriously benefit from in this rework. 
  - i do not think putting the temperature gun in the armory is a good idea because of this. science needs something unique and powerful to gun for with their T3s, and i genuinely can't even remember arsenal's other T3s right now. not implementing the temperature gun as an arsenal tech would just make the issue far worse. 
  - if lasers began using energy cages, this would be a very easy way to make the arsenal tree something appealing for science to go for, both during times of crisis and during normal play. helping security directly buff their equipment would feel worthwhile, and it creates a fun choice between civilian power cells and military power cells. i think self-charging energy cages would be a good T3. energy cages also just make more sense than sticking an entire rifle into a battery charger.
- the ion rifle should be a science tech. i understand that IPCs are not very respected due to how powergamey they are, but this would pretty much instantly disable the ability for IPCs to ever be antags. having a gun that specifically targets one species at roundstart that can literally two-shot them is just way overtuned, in my opinion. not to mention its interaction as an EMP, which would potentially cause fuck ons of collateral. the only way antag IPCs would be able to function is if they immediately raid armory and steal the ion rifle, which is putting a stupid amount of pressure on them for very little reason.
- lasers should not be projectile-only. they already shoot slower and for less damage than ballistics, they don't also need to be extremely hard to aim and easy to dodge. things like the energy shotgun should still be projectiles, but that's explicitly due to how they have a spread-out bullet pattern, which would be absurd to play around if it was hitreg. 
  - the energy pistols are perfectly fine as projectiles, this is mostly in reference to laser rifles and cannons.
 

things that have been said that i do not agree with are fair criticisms: 
- nukies creating nothing but burn chems doesn't actually make much sense. crew has tons of ways of doing pierce/blunt/slash outside of guns, and stuns can still very easily result in guns swapping hands. if the nukies choose to exclusive go with DEswords, then this might be a fair point, but that opens them up even further to stun meta.
- the HoS already has a unique energy gun. you could tune it up, but removing the WT does not actually change much.
- onboard antags not being able to instantly heal off the damage they take during security fights by cutting up clothing is a good change, not a bad one. it actually incentivizes buying and using syndicate meds, rather than just raiding the clothesvend. 
  - if you really want to, you can just steal ointment from med. it's really not that hard.
- crew already rarely uses alternative damage types during nukie rounds. nukies usually just buy energy shields and stack pierce resist. and that is more than enough, most of the time; nukies rarely go down within the first, second, or even third fight with crew. it's only after their meds have been whittled down and they repeatedly get caught out that nukies usually get punished and die.   
- lasers are already the best weaponry against nukies. you just need to utilize them in a semi-smart way. nukies can blast through windows, but you're still constantly whittling them down. they can stack as much heat resist as they want, but their heat resist is explicitly getting nerfed within this design doc. i don't think it's that big of a deal.
- security is not meant to be able to gun down hordes of people. requiring that they retreat to recharge their laser weaponry is perfectly fine and makes much more sense in-lore. 


misc thoughts:
- i generally agree that more sci-fi guns is fitting and more fun than the standard ballistics.
- i think security having their strongest guns roundstart (as is currently the case) is very bad for the game. this doc would hopefully change that. 
- i like the fact that melees, slips, etc. are much more heavily emphasized under this doc. they feel borderline pointless if you're literally fighting for your life as a random crew member. giving syndies a reason to punch the lights out of secoffs with northstars is fun.
- being able to get away from sec using EMPs is fun. it incentivizes not trying to box out the entire security force at once, and instead making getaways. 
- harm batons would probably be used more. based
- eswords would probably be used more. based
- this would definitely require a second server or something to test on first. no way this could be introduced cold turkey, even with a monthlong rollout. extensive testing would need to be done before it could be considered for main server play.

i think that's everything. some of these might be rambly or incoherent but i wanted to be as comprehensive with my thoughts as possible. hopefully it helps.

The048:
For one, I disagree with the concept of making lasers projectile-based entirely. Lasers are quite literally beams of light, one of their fundamental traits is that they have no travel time. If the new projectile-based weapons were to be flavored as plasma or just general energy weapons I would be more ok with it, but making the laser weapons blatantly unfitting of their concept would be dumb. I would personally rather have lasers stay as weak but easy to use weapons designed for mass use in disaster scenarios as opposed to being functionally the same as the main weapons used by security.

However, my big issue is that this would fundamentally destroy a lot of the effectiveness of security, especially in the removal of magazines for standard weapons. I've heard some speculation on the idea of researchable magazines, but honestly that should be available by default for the standard and miniature energy gun, since 12-15 shots of a projectile energy weapon isn't enough for any kind of significant combat, especially given how strong most antagonist weapons are in comparison. At the same time, I do like the idea of having researchable magazines for other types of energy weapons like the laser cannon, but in the case of standard-issue weapons, having magazines is non-negotiable in order to make security still functional beyond green alert.

That leads to a big issue of this: antagonist balance. Simply put, switching security to be energy-focused would take a LOT of rebalancing on the part of antagonist equipment. The amount of armor that would be a problem is already huge, but also most antagonist weapons would need pretty significant nerfs in order to not completely steamroll security, unless the energy weapons were to be significantly buffed from their current state on other forks. Since a big part of this seems to focus on the idea of curbing power creep, I assume that antagonist weapons would be nerfed in kind, but it is important to note that this will be a pretty big undertaking.

Lastly, I don't think the multi-week rollout planned here is a great idea as is. For one, there is no mention of a wholistic balance reevaluation, which means that if this were to go as planned the game would be entirely antagonist-sided in all respects. Secondly and more importantly, I don't think it's a good idea to roll this out entirely in pieces. I think it would be better to simply roll all of the changes at once, then monitoring for necessary tweaks.

I am not opposed to this idea entirely, but I do think that caution should be exercised when approaching it.

src/design-proposals/lasers.md
```diff
@@ -0,0 +1,154 @@
...
+
+The removal of the WT-550 from the Head of Security office.
+
+The removal of disablers from Sec officer lockers.
+
+Overall, Armory would now contain:
+- 1 disabler safe (containing 5 disablers)
+- 3 energy guns
+- 3 laser rifles
+- 3 Kammerers
```
taydeo, line 67: Kammerers? Why?

src/design-proposals/lasers.md
```diff
@@ -0,0 +1,154 @@
...
+
+Kammerers should stay as they are explicitly very CQB weapons, as opposed to the longer-range, more reliable lasers.
+
+The addition of a disabler safe to all armories for emergency Command-arming reasons.
+
+The removal of the WT-550 from the Head of Security office.
+
+The removal of disablers from Sec officer lockers.
+
+Overall, Armory would now contain:
```
taydeo, line 63: The entire roster makes security look pretty lightweight, which, while it does align with the sort of "Power-tripping mall cop" status we want security to embody, may potentially leave them woefully under-prepared for some basic station events. For example, what would be of the station if a dragon came in with this sort of load-out? Are the Kammerers listed below the solution for this? And wouldn't that be kinda weird? Why does the Warden get to carry a laser shotgun instead of replacing the Kammerers with the laser shotgun, and just giving the Warden a different weapon all together?

src/design-proposals/lasers.md
```diff
@@ -0,0 +1,154 @@
...
+See https://github.com/space-wizards/space-station-14/pull/35447 for implementation. Main difference: this will be an Armory weapon and not researchable.
+
+### Ion Rifle
+A powerful three-shot rifle that deals 50 ion damage on a direct hit as well as a 3x3 tile weak EMP where it lands. Very powerful against cybernetic threats, such as borgs, IPCs, or mechs, but only tactically useful against other energy-weapon users.
+
+### Energy Gun
+A battery-powered carbine (can fit in the backpack) which can fire either disabling or laser bolts, with less charge than the base Laser Rifle. It has 15 charges.
+
+### Miniature Energy Gun
+The replacement for both the Officer's disabler and pistol, this is simply a smaller version of the Energy Gun. It can fit in your pocket and has 12 charges, which can be used to disable or laser.
```
taydeo, line 45: Big fan of this, and would greatly help with immersion for both new cadets and already experienced sec players.

src/design-proposals/lasers.md
```diff
@@ -0,0 +1,154 @@
...
+
+- Laser projectile changes, damage tweaks to all current laser weaponry.
+- Minor (-10%) nerf to heat resist on all nukie suits.
+- Initial rollout of Miniature Energy Gun.
+- Addition of the Miniature Energy Gun to Cargo.
+- Miniature Energy Gun will replace Sec Officer's pistol and disabler, as well as the HoP's disabler plus.
+- Initial rollout of the Energy Gun.
+- The addition of Energy Gun crates to Cargo.
+- Replacing the Enforcer crate in Cargo with a Kammerer crate.
+- Removal of the SMG crate in Cargo.
```
taydeo, line 132: Are we sure with a complete removal? Why not incredibly expensive?

src/design-proposals/lasers.md
```diff
@@ -0,0 +1,154 @@
...
+- Changing all laser-based weaponry from hitscan to projectile, except for the retro laser blaster because it is funny.
+- Remove ion damage from disabling bolts.
+- Reduction of the total amount of disabler shots-per-charge.
+- Reduce the reflective vest's deflect chance from 100% to 66%
+
+### Possible (Future) Tweaks and Concerns
+
+New Arsenal research paths for better energy / ballistic weaponry, printable by Sec to make up for the reduction in overall roundstart firepower.
+
+Certain antag's heat resistances, most probably Nukeops and Dragons, will likely need to be tweaked down a significant margin.
```
taydeo, line 96: This makes sense and partially addresses my concerns above.

src/design-proposals/lasers.md
```diff
@@ -0,0 +1,154 @@
...
+
+### Technical / Misc. Tweaks
+
+- One unified Laser Projectile and one unified Disabler Projectile defined in Funkystation files.
+- An overall increase in laser damage.
+- The removal of self-recharging from the HoS' multiphase gun (why is it a better captain's laser?)
+- The removal of self-recharging from the EG-4 energy revolver (Blueshield)
+- Reduce the Warden's energy shotgun to 6 shots.
+- The removal of the mk58 pistol and disabler from the Security Officer loadout for the miniature energy pistol.
+- Changing all laser-based weaponry from hitscan to projectile, except for the retro laser blaster because it is funny.
```
taydeo, line 87: This I see being a big point of contention, and while I understand it doesn't make much sense, I think it's for the best ultimately because of the sci-fi tropes built around this unrealistic yet visually stunning design decision made in early sci-fi work. It greatly changes the experience of both fighting against laser weapons and using them, in a positive way.

src/design-proposals/lasers.md
```diff
@@ -0,0 +1,154 @@
...
+- Replacing the Enforcer crate with the Kammerer crate.
+- Replacing the smg and pistols crates with the Energy Gun and Miniature Energy Guns crates.
+
+### Technical / Misc. Tweaks
+
+- One unified Laser Projectile and one unified Disabler Projectile defined in Funkystation files.
+- An overall increase in laser damage.
+- The removal of self-recharging from the HoS' multiphase gun (why is it a better captain's laser?)
+- The removal of self-recharging from the EG-4 energy revolver (Blueshield)
+- Reduce the Warden's energy shotgun to 6 shots.
```
taydeo, line 85: Please review my comments about the energy shotgun and revisit this.

taydeo (changes requested):

Terkala:
Removing so many ballistic options from the crew would make almost all forms of nukie/traitor armor obsolete, except for the elite and jugg suit. Which are already basically meta for effective armor due to Funky having extra laser weapons available to the HoS and BSO.

While I agree with the thematics, balancing around this would be incredibly difficult.

src/design-proposals/lasers.md
DevilishMilk: I believe the pump shotguns should still have their niche as the Security ballistic weapon. The reason I'm so vehemently opposed to current ballistics within the Armory is that full-auto weaponry, just by its nature with our combat system, is generally going to be incredibly oppressive and balancing around it will either lead to them feeling:
a) inauthentically weak (why are people just bullet sponges)
b) authentically strong (okay, this is just objectively better than everything else, so why would i use anything else?)

The Kammerer, being a shotgun, doesn't necessarily have the same issues within combat as the Drozd and Lecter do due to it requiring much closer-range use, as well as not being able to keep shooting as long as ammo is had. Time needs to be taken to reload, and that's time that the other side has to either get closer and take someone down, or run and lick their wounds.

Antag balancing concerns here regarding hitpoints and armor values can always be adjusted as needed, as I've stated a bit further on down in this PR I believe it will be counterintuitively easier to balance the Armory and antag armor values going forward should this doc be implemented.

I also think purely from an aesthetics standpoint that the laser shotgun is a pretty neat item, and it should be unique.

src/design-proposals/lasers.md
DevilishMilk: The SMG as-is is incredibly oppressive (even in current balancing, where it's pretty much just Drozd+Lecter but better) which is what I'm currently trying to completely steer clear of regarding ranged weaponry, and the other main intent of this design doc is to move ballistic weapons, primarily full-auto ones, over primarily to the Syndicate. I don't think cargo budget is a good hurdle to restricting access to certain things as it can be pretty quickly cheesed.

DevilishMilk:
Thank you to the folks with feedback, I got real busy with work real quick and ended up not having much energy to do anything on this front. I'll address some of the points brought up:

Do I think that the full implementation of this will necessarily shake our combat system? Not necessarily; as with any other change there'll just be a new meta. My argument here is that the current meta, which is just roundstart-available best-in-class ranged weaponry, is unhealthy for the game as it stands, and this PR aims to create a healthier foundation for these changes you've highlighted to be made.

I can argue stun-meta here but ultimately I think it's irrelevant to the design doc as is, and also believe that that is something that should be addressed elsewhere (like the stun resistance changes that you, ferynn, ported from Wizden IIRC? They're a step in the right direction IMO) and can continue to be done so.

As for the concerns about the lack of reloading and overall less of an ability to sustain a firefight, that is intentional and I think an integral part of this doc. The ability to just carry stacks and stacks of ammo in your inventory to continue shooting I think just makes combat an attrition game rather than a cat-and-mouse back-and-forth game which is generally (in my Opinion) more fun. I have made some possible research items in the design doc, one from 13 (the Inducer) and a couple of silly weapons for more variety.

I think the HoS' unique multiphase gun is enough in and of itself as an HoS weapon. Making them a walking arsenal just seems kind of silly for what we're going for.

PR #12, opened 2025-07-05
The Distress Beacon, and Beacon Responders

jackel234:
My design doc... it is finish...

I'd be happy to talk about the ideas in here, and if it gets accepted, even happier to have help making it. I'm good at writing my ideas out, but actually implementing them is another thing. I also only included on example for NT-aligned, Neutral, and Hostile, mainly because I didn't want to be working on this design doc until the end of time. If anyone has ideas, feel free to share them, especially since this idea'll need a lot of possible scenarios to actually work properly.

### DIFF
--- a/src/design-proposals/distressbeacon.md
+++ b/src/design-proposals/distressbeacon.md
@@ -1,0 +1,167 @@
+# The Distress Beacon, and Beacon Responders
+
+| Designers          | Implemented | GitHub Links |
+|--------------------|-------------|--------------|
+| cattyowo/jackel234 | :x: No      | TBD          |
+
+## Overview
+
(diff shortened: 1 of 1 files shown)


jackel234:
> love this idea!! it solves a couple of issues we've noted before, such as there not being a lot of ghost roles. i particularly love the roleplay beats you outlined, and i agree they could genuinely add to the round!
> 
> my (very minor) concerns are:
> 
> * is there anything to discourage players from calling a distress beacon during calm/greenshifts? or should this be something admins should enforce (if it's okay or not to call a beacon)
> * how do you expect the pool of possible events to be weighted? like, are the responses going to be primarily helpful, or do you want them to lean more funny?
> 
> both questions are very minor, i think this is a good idea!!! thank you so much for your work :-D

thank you!! :)
as for your questions:
i think it'd mostly be something the admins have to enforce. i could make it so the buttons that unlock the beacon can only be presed during certain alerts, but i think that'd be a bit too much of a challenge for me, as well as potentially being messy. the main thing that would hopefully discourage misuse if the fact you need both the hop and captain to agree to pressing it, but i can see potential of that not being a huge barrier.
generally, i'll have the events weighed more towards neutral responders, which since most of the comedic responders are neutral, it'll likely lean more towards funny, rather than truly helpful.

PR #14, opened 2025-07-05
Fear And Death RP Design Doc

UristMcWiki:
[DeathRPForFunky.md](https://github.com/user-attachments/files/21083103/DeathRPForFunky.md)

### DIFF
--- a/src/design-proposals/DeathRPForFunky.md
+++ b/src/design-proposals/DeathRPForFunky.md
@@ -1,0 +1,30 @@
+Urist McWiki |	Partially/Open PR | TBD
+
+Overview:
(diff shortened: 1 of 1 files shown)


corresp0nd (commented):
while i agree we do have an issue with people validhunting, and focusing too much on winning or losing, i am iffy on this particular solution. the implementation of this solution, in particular, seems very vague. i don't particularly understand how these changes will actually change our server culture. 
in particular, these seem like balance changes in order to discourage a certain type of play style rather than new systems to *encourage* better roleplay. it particularly seems like this proposal favors RRing more, which is explicitly something we tried to move away from. i agree that round removal can be healthy for the game, and can add to a shift, but it doesn't make sense for us to try and return to that standard when none of the initial problems with RR have been resolved (people can do nothing except close the game or observe if they get RR'd). 
i agree that there are certain actions and play styles that players can favor that do not have a place in an mrp environment, but overall, this proposal (and honestly, a lot of your recent changes) have felt like they are punishing people for not playing the "right" way. it's fine if people are good at the game, whether that be through the medical system or through combat. we should not punish people for "solving" a system and figuring out the best optimized way to do something, we should be improving those base systems and trying to encourage people in directions that may not be optimal.

PR #18, opened 2025-07-10
Chaplain overhaul

Teasq:
This is a document discussing potential changes to the chaplain, hopefully giving it a more clearly defined role on the station while enriching roleplay.

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -32,13 +32,12 @@ Proposals
   - [Map Template](design/maps/template.md)
 
 - [Game Design Proposals]()
-  - [Hydroponics Changes](design-proposals/botany-changes.md)
-  - [Command SOP](design-proposals/command-sop.md)
+  - [Hydroponics ](design-proposals/botany-changes.md)
   - ["Corporate Secrets"](design-proposals/corporate-secrets.md)
   - [Lasers and Armory Overhaul](design-proposals/lasers.md)
-  - [The Metashield](design-proposals/metashield.md)
   - [Company Scrip](design-proposals/scrip.md)
-  - [Security Training Manual](design-proposals/security-training-manual.md)
+  - [Medical](design-proposals/medical/medical-direction.md)
+    - [Scanners & Triage](design-proposals/medical-scanners.md)
 
 - [Map Proposals]()
 
--- a/src/design-proposals/medical/medical-direction.md
+++ b/src/design-proposals/medical/medical-direction.md
@@ -1,0 +1,45 @@
+﻿# Medical Department: Direction & Overview
+
+| Designers  | Implemented | GitHub Links |
+| ---------- | ----------- | ------------ |
+| correspond | :x:         | TBD          |
+
+## Overview
+
+This is meant to be a guiding document for the entirety of the medical system and department. These can overall be summarized into design principles, however, some outlining systems are included. The principles themselves take precedence over any example systems listed in this document, as it is meant to be a guiding direction rather than a strict timeline. If you have any questions about this document, feel free to discuss them in the medical workgroup or contact @correspond on Discord directly.
+
+## Design Principles
+
+### I. Suspend Your Disbelief (in the Name of Slapstick)
+
+Above all else, the primary principle to consider for medical is the *aesthetic* of it. The why or how we get there don't actually matter that much, the primary goal is that there is a story to be told through the game. When players get hurt, they should get hurt. Simple as that. But the ways we flavor this depend on the primary goal of each fork.
+
+For us, Funky Station, we're looking for a world that takes itself serious, even in the face of silly interactions. The best way to do this is to create a system that represents real biology just enough to push it into the suspension of disbelief; by backing these situations with science, we can push them to contemplate even the fantastical.
+
+Similarly, though, the systems that we are basing this world off of are interesting! Okay, sadly, I am not a STEM major or doctor or anything. But genuinely, biology and all of its subdisciplines are really interesting and seeking to emulate those very interesting parts is a primary goal of the system. The aesthetic we are seeking, on top of this seriously-silly slapstick, is a genuine interest and passion for the systems being created.
+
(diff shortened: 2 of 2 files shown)


Midna-Midnight-Miami:
I helped with this colon three

src/design-proposals/chaplain-overhaul.md
```diff
@@ -0,0 +1,99 @@
...
+|---|---|---|
+| Teasq | :x: No | TBD |
+
+## Forewarning
+These changes are unlikely to happen anytime soon, so please make balancing suggestions with that in mind. As of right now, I am planning on rebalancing the gauntlet and all the other nullrods. With the increase in effective "chaplains", I will also obviously make stuff like closing rifts take much longer and potentially damaging to the acolytes. Number discussions are welcome, but I don't want that to be the predominant discussion until the ideas are set in stone.
+
+## Overview
+This document will propose several ideas I've been thinking up for quite some time regarding the chaplain's role on the station, but the key points are:
+
+• Chaplain will become an authority on magic/the occult, similar to the magister, with one predominant role: enforcing what magic is legal/considered evil and what religion is sanctioned aboard the station.
```
corresp0nd, line 13: really love this idea! i think it suits the chaplain really well and fills a hole that has been kind of iffy for a while. the only thing we should consider is how much power we want to give chappy, esp bc they are not mindshielded

src/design-proposals/chaplain-overhaul.md
```diff
@@ -0,0 +1,99 @@
...
+The reason for this being that there currently isn't an authority on magics as of the removal of EOC and navigating magic use in general has been confusing for many.
+
+• Similarly to the nullrod, chaplain will - preferably roundstart - select a sect. Sects will be themed around the already existing altars. This sect defines the flavour of whatever religion they decide to practice this shift. 
+More details on sects below and will be expanded as I get more ideas/feedback
+
+• Chaplain will have a small group of acolytes, likely 1-3, who follow the chaplain and whatever sect they decide upon. They will be notably less capable than the chaplain,
```
corresp0nd, line 19: a little bit iffy on this but im not averse to it... i think id have to see how this is implemented to give a better review of it ^^;

src/design-proposals/chaplain-overhaul.md
```diff
@@ -0,0 +1,99 @@
...
+
+• An actual SOP for the chaplains and acolytes to define the role of NanoTrasen's religious sect more clearly.
+
+• Debating whether or not nullrods should be attached to respective sects
+
+
+## Features to be added
+• Various mapping changes to the chapel as a whole to accommodate 1-3 extra acolytes depending on map size and intended population.
+
+• Addition of the acolyte job.
```
corresp0nd, line 30: maybe it would be better if the chaplain could pick their own acolytes....? and ofc get permission from hop/command as well. or maybe it could be a randomly rolled crew subrole?? much to think about

corresp0nd (changes requested):
love the idea, and i do agree chaplain could use the love!! 

main things im iffy on:
- i dunno if acolytes should be their own unique *role.* i am open 2 change my mind but i think theyd be better as something else
- miracles should be a once-per-shift, if even that, type of thing. im worried that chaplain mains will figure out ways to speedrun getting them every shift if we make it like, a bar, or something similar :p

src/design-proposals/chaplain-overhaul.md
```diff
@@ -0,0 +1,99 @@
...
+• Crafting for altars in the case that a chaplain wishes to follow a sect that doesn't match the roundstart altar.
+
+
+## Acolyte
+• Requires a PDA. I'll probably resprite the chaplain PDA at the same time as making this one.
+
+## Chaplain and acolyte SOP
+WIP
+
+## Sects - WIP
```
SigmaTheDragon: An idea for a sect:
Stars (Cosmic Altar):
- Bible: Deals cold damage on fail, can heal asphyxiation damage unlike normal bibles

- Miracle(s): May be able to gain some abilities from the cosmic cult, such as eschew metabolism, void step and even the nova launch thing as their ultimate and most costly miracle

Maybe to gain progress to the miracles they have missions like finding 'stars' around the station (similar to the malign rifts but with unique sprites and only acolytes/chaplains can see, also don't change the terrain around them)

src/design-proposals/chaplain-overhaul.md
Teasq: This was before corporate secrets was fully introduced. I don't entirely understand how that works yet, but I'll review both that and this document and get into gritty details as I set up my workspace (I've recently swapped to Linux and there's a whole learning curve to it).

With that said, at least for now my current reasoning is 1) they don't know shit about the corporate secrets and are genuinely brainwashed that their clerical magic is objectively "good" and all other unsanctioned magics is evil, and therefore would actually know how to operate in the cases of a cult antag, etc. They can't roll cultists, so I think them being the "authority" on the 1/1000 cult games where they also happen to be an antag is interesting.

And 2) maybe they could function similar to the IAA for magic exclusively. They can find and point out what they consider to be illegal magics and defer to the higher authorities to deal with it. I think it's a decent compromise between giving them authority on magic while still keeping them unmindshielded. Much like the NTR, this relies on people actually RP'ing properly but it's a MRP server so I don't see that being an issue generally.

corresp0nd (approved):
looks goodddd

src/design-proposals/chaplain-overhaul.md
Teasq: Hey, I'm only just now getting back on this so sorry for the big delay in my response. This isn't really the goal I have in mind for chaplains, as it's overall too powerful and too "gamey". 

My wish for miracles isn't that it is some kind of in-game progression system, but rather that it is a tool that strongly supports roleplay and the fantasy of the sect that it is tied to. That means the miracle generation, regardless of sect, is a set resource that is easy to balance around via making more "useful" miracles much more expensive and limited.

PR #20, opened 2025-07-19
Salvage rework poposal 

ThatOneGuy227:
A proposal to partially rework salvage in order to more clearly define its role as a resource provider for the station,

### DIFF
--- a/src/design-proposals/Salvage rework proposal.md
+++ b/src/design-proposals/Salvage rework proposal.md
@@ -1,0 +1,113 @@
+## Salvage Rework Proposal 
+
+| Designers | Implemented | GitHub Links |
+
+|---|---|---|
+
+| ThatOneMoon| :x: No | TBD |
+
+## Overview 
+
+A proposal to partially rework salvage in order to more clearly define its role as a resource provider for the station,
+While also much more appropriately rewarding the salvagers who do so,
+As such, this document has three primary proposals. 
+
+## Salvage Tickets 
+
+Salvage Tickets would be a very straightforward solution for the problem of salvagers not focusing on resource gathering,
+inspired on how it exists in SS13 and certain 14 servers, when salvagers process raw ores or scrap, they also get salvage tickets.
+
+These tickets can then be exchanged inside a special vendor for items and equipment salvagers want, such as medical supplies, tools, weapons and suits.
+
+The vendor would not be able to be moved and taken aboard the salvage ship, meaning in order to use it, they have to return to cargo to redeem tickets for equipment,
+In doing so, gathering materials and getting equipment are something that go hand in hand. 
+
+ideally leading to scenarios where salvagers will first go safely gather materials and process them, 
+and use the tickets they gain to get the equipment that will allow them to take on more risky locations. 
+
+### Features to be added
+
+## Salvage packages 
+
+In short, Salvage Packages are nearly the same as regular mail packages that cargo receives from the mail teleporter.
+
+But they have a few key differences: 
+
+- Salvage packages are only ever found from salvage itself, spawning as part of wrecks, ruins and debris. 
+
+- Instead of being addressed to a specific individual like cargo mail, salvage packages are addressed to a department, 
+  and anyone from that department is capable of unlocking and opening it.
+
+- Unlike cargo mail, packages primarily contain items related to the department it is intended for.
+ 
+ (Meds for medical, food/ingredients for chefs, mutated seeds or hydro chems for botany and so on)
+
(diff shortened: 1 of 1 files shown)


src/design-proposals/Salvage rework proposal.md
```diff
@@ -0,0 +1,113 @@
...
+and allows departments to occasionally get their hands on some more exotic items. 
+
+(Such items could even be low-mid level contraband for added spice.)
+
+## Mining/Scraping Rework and Rebalance
+
+A set of changes to attempt to unify mining and scraping mechanics for the sake of streamlining the process. 
+
+- The ore processor is no longer used to refine ores; instead, now both ores and scrap get processed in the recycler (renamed into the material processor). 
+  [Due to the recycler not working properly with stacked items, a rebalance of ores will be required to account for them being non stackable.]
```
corresp0nd, line 62: you can edit the recycler system to be able to do this! its just that stackable items is a component appended onto entities, you need to add a case for what it needs to do to handle entities with this component.

src/design-proposals/Salvage rework proposal.md
```diff
@@ -0,0 +1,113 @@
...
+this helps address the issue of station crew feeling like salvage is not working, or are detached from the station,
+and allows departments to occasionally get their hands on some more exotic items. 
+
+(Such items could even be low-mid level contraband for added spice.)
+
+## Mining/Scraping Rework and Rebalance
+
+A set of changes to attempt to unify mining and scraping mechanics for the sake of streamlining the process. 
+
+- The ore processor is no longer used to refine ores; instead, now both ores and scrap get processed in the recycler (renamed into the material processor).
```
corresp0nd, line 61: the ore processor has genuine utility as being able to store entities, plus having the ui is generally pretty nice. is there a particular concern as to why we need to make them one object?

src/design-proposals/Salvage rework proposal.md
```diff
@@ -0,0 +1,113 @@
...
+(Such items could even be low-mid level contraband for added spice.)
+
+## Mining/Scraping Rework and Rebalance
+
+A set of changes to attempt to unify mining and scraping mechanics for the sake of streamlining the process. 
+
+- The ore processor is no longer used to refine ores; instead, now both ores and scrap get processed in the recycler (renamed into the material processor). 
+  [Due to the recycler not working properly with stacked items, a rebalance of ores will be required to account for them being non stackable.]
+
+- Ores or scrap items that are thrown into the recycler now also print out a ammount of salvage tickets
```
corresp0nd, line 64: look at how goob does this for lavaland; afaik you put ores into the processor and it generates tickets based on that.

corresp0nd (approved):
this are good changes to help encourage salvage to be a little bit better! i am only minorly worried about some of the implementation, but it's not a suuper big deal

src/design-proposals/Salvage rework proposal.md
ThatOneMoon: I actualy did!
And i also did not like how their implemintation looked, rather instead i am going to base it more on how Starlight does it
But the thing both Goob and Starlight do that i dont like is that tickets are *only* printed from mining, meaning the salvs that like to gather resources from scraping are not actualy getting tickets for resource gathering

src/design-proposals/Salvage rework proposal.md
ThatOneMoon: Yep i know that, my current WIP of the mining rework already has this implamented, required a bit of fiddling but its working how i want too now

src/design-proposals/Salvage rework proposal.md
ThatOneMoon: A number at least to me
The first is that while the UI is nice, any person that has worked long enough in salvage can tell you using the Processor gets kinda quickly annoying due to you having to spam click to make the resources 

Another minor issuee is that salvage borgs actualy cannot even use the ore processor because they cannot insert ores into the machine 

And in my opinion the bigest problem, the Industrial Ore Processor in its entirety and why i plan to remove it (and not give any kind of replacemnt) 
A very, very, very common issue is that salvs that do actualy like to mine ore sometimes dont, because they are just waiting for the industrail to be gotten, meaning they end up not delivering resources to the station because they want to be "efficient" 

Its also just too powerfull in general, the ammount of extra resources you can get from it is frankly insane and is basicaly a headache to balance around, especialy in regards to salvage tickets being introduced 

giving only a single baseline of resource return rather than 2 drasticly diffrent ones is both more balanced, and easier to work with

PR #24, opened 2025-08-22
Science change proposal

JoulesBerg:
Hi. I was bored, and so I made this.
Just a blah of ideas rn, and as I said I don't really play science because I think its dumb, but these are my ideas to make it better. It is fully possible they are naive / I don't fully understand the current system well enough to make some of the comments I made, but thats why this is a proposal. I appreciate any feedback

### DIFF
--- a/src/design-proposals/science-overhaul.md
+++ b/src/design-proposals/science-overhaul.md
@@ -1,0 +1,117 @@
+# Research Point Types (Science Overhaul)
+
+| Designers | Implemented | GitHub Links |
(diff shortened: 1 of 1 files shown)


JoulesBerg:
My thought is just a generic points exchange, so if you are solo you can pick what you are BEST at, and just go hard. The exchange rate would be like, 2 to 1, so very inneficient, but it covers the base of "well shit, we have zero access to this thing we need to x research so guess thats just not happening"

PR #33, opened 2025-12-06
IPC Replacement: Reploids

ParasiteSaintLeech:
Design doc for Reploids.

Baby's first repo + branch; I hope I did this right.

There's too many issues with IPCs. There's too many issues with the way they are balanced. There's too many issues with how they interact with opposition, space itself, and lore. For for whatever reason, people also don't like that they can't eat(?). 

This addresses all of that and more, and maybe saves the original concept, too. Besides, if this goes through, I get to do some old concept characters more accurately and maybe do some spritework.


ParasiteSaintLeech:
> ferynn

Addresses this now.

src/design-proposals/reploids.md
```diff
@@ -0,0 +1,147 @@
...
+- Just for fun: Playing MIDIs.
+
+Most likely in the same vein/instrument as pAI MIDIs are played. Maintaining that 'artificial' vibe.
+## Game Design Rationale
+
+This kind of thing seems as if it'd be a no-brainer for the kind of sandbox stuff Funky seems to want to provide. Not only that, there ends up room for quite a bit _more_ than what was bargained for initially, as it's no longer 'just a robot' and etends into bio-organic territory, too. I'm unsure of what more one would want to add mechanically; but that can be fleshed out in the future. Making the basic concept exist is the most important part; all the mechanical stuff can be fleshed out later, if that is more beneficial.
```
corresp0nd: not entirely true: while we do want players to be able to create a story, it doesn't inherently have to be in a pure sandbox environment. think of it this way: the game itself is a dungeon master for a tabletop game, the players will make their own choices, and the dm will take into account the players, but the dm ultimately has the final say on what the world is like and what is considered realistic within that world. players respond to the world, not necessarily molding the world to fit themselves.

corresp0nd (changes requested):
i've set this to request changes, but i do think this is a really good foundation for how the niche of ipcs can be filled. so, keep working on it, think of it less as a "no" and more of a "the middle is still pretty raw, we can bake this for longer" 

reploids are a pretty cute idea. they still have that synthetic niche that we want without actually being completely silicon. i like the idea that they're artificial flesh with metal moving parts, it's a cute way to keep them synthetic while still being organic. i particularly love the bit you added about them not having any sort of immune system at all, it's a fun reason as to why they can be modified so heavily without any fuss. it also has the bonus of leading to possible later game systems (they would be great for viro!). 

the main thing im worried about is that we don't actually want them to just like, be able to look like other species. this is one of the mistakes we have made prior and we don't want to repeat this. reploids should have a distinct visual style, and while being heavily modified can be a part of that, easily being able to just Look the exact same as another species is completely contradictory. 

the second thing i want to bring up is the connotations of a species being artificially made for the exclusive purpose of being a worker. i am not telling you to cut this part of their lore, really we welcome any sort of critique of capitalism, but this is not something we can tack on without careful consideration. i do not want anything that can promote discriminating against fellow players because of lore (think 13's infamous "definitely not slurs" towards lizards and our general policy on the word "clanker"). 

overall, this is a really good direction, and im excited to see where you take it!

src/design-proposals/reploids.md
```diff
@@ -0,0 +1,147 @@
...
+  `Poison: +50%` [Alternative idea; make all chems metabolize slower]
+  
+  `Radiation: +50%`
+
+  
+
+`Native' Traits: These are **potential** mechanics that might be added to them in order to give them some flavor.
+- Damage Assesment
+
+Using an action, accurately deduce the state of one's internal, if they are sick, what is hurt. Basically, personal vitals sensors. This leans into them being artificial, and additionally being fully aware of their own artificial nature.
```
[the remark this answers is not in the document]
corresp0nd: sort of hard to say? we had a trait called "self aware" that was like this and i ended up hating it. it encourages people to ignore med and just heal themselves completely, which we generally don't want

corresp0nd:
> I do apologize for 'baking in conflict,' too; I'm used to priming settings for conflict in my writing. I suppose such is not needed in Funky!

doing this is perfectly fine, tay particularly loves to try and add class conflict to a lot of their prs (this is the point of scrip, unions, and the direction we're taking revs). i only brought it up because we've had some issues before with specisism and i don't want it to seem like we're encouraging players to keep doing that. it's a hard balance to keep in mind :P

JoulesBerg:
In general. I think the concept of "but also" should be avoided for uniqueness and flavor.
The specific thing I mean in this case is their ability to consume oil, lube, etc to satiate thirst, but also they can just consume normal stuff also.
Or the proposal that food would also satiate thirst for them, but at a cost of making them slow or something iirc what you wrote.
Both of these are just straight buffs to the species and / or require new species specific implementations / modifers, while not necessarily flavorful or additive to the species uniqueness. IMO they fall in the same lore black hole that cannibalizing species markers falls in. 
I think it IS cool and flavorful that they drink synthetic compounds like lube, oil, or welding fluid. But you immediately ruin that flavorfulness by turning around and saying "but if they wanna drink water or milk thats cool too".
Its like trying to PR a change that allows moths to eat normal food. Why? All it does is REMOVE flavor by running it back like that.
Implementing the food satiating thirst thing idk how hard that would or wouldn't be. I forget how the two systems work. But I don't think it would work out of the box, and again, I have to ask the question of "Why is it flavorful for them to be able to satiate thirst by eating food?"
It doesn't really add anything, and detracts from the cool and flavorful idea that they can only satiate thirst with synthetic compounds. Because when I can't find welder fluid conveniently I can just go rob the chef.

There were a couple other things I felt like asking "Why though?" But my general take is theres probably too much stuff in this right now thats half baked (dough in the center) or trying to be too universally appealing, which defeats the purpose of a species imo. 

It should be unique, and not necessarily universally appealing, whether because of some significant drawbacks, or whatever.

My personal philosophy would be to take a base species like human, and overall nerf it. Maybe some bonuses here and there to offset SOME bad stuff, but overall the species should be slightly *worse* from a pure balance perspective, because then the people that play it will play it for the right reasons.

Thats my take anyway.

corresp0nd (approved):

src/design-proposals/reploids.md
```diff
@@ -0,0 +1,134 @@
...
+
+  
+
+`Native' Traits: These are **potential** mechanics that might be added to them in order to give them some flavor.
+- Situation Appraisal
+
+Using an action, deduce basic info on internal an expert climates. This will pull a UI element that tells the player the temperature, if the oxxygen level is too low, their hunger/thirst in a % value, and the amount of blood they have at that moment.
+- Emag interaction
+
+Becoming Emagged causes reploids to experience auditory and visual hallucinations, and induces the 'drugged/drunk' filter on ocassional intervals. They would begin to twitch, visually. Potentially fun things might additionally be cuasing the combat mode to toggle at random. The "total" time that this occurs should be limited, somewhere between five and fifteen minutes. The idea to is impair them, not to completely debilitate them for the rest of a given round.
```
TheHolyAegis: Can I ask what this means? "Twitch?" If it's the screen flickering, then you really need to keep in mind with people who are sensitive to flickering lights

ParasiteSaintLeech: Does this satisfy the review?

ParasiteSaintLeech:
Jeopardy Theme Plays Louder

PR #39, opened 2025-12-29
Create  Vitality-Based Medical and Cybernetics System

Terkala:


corresp0nd (approved):
looks great!! im pretty excited to see how this goes, i do particularly love the modular aspects you outlined. my questionsssss:
- do we reaaaally want skills? weve discussed them before and usually the answer was no, not really, but i can still be convinced. or we can instead just tie all of the medical benefits you mentioned to either the tools or a specific machine (i guess like, the body scanner machine in 13? whatever its called). im leaning towards a machine tbqh, i feel it would establish the same end result w/o skills
- how visible should integrity be to players? its sort of described as another health bar here (though slightly abstracted in the way ttrpgs tend to be), and im not super set on the idea of it *just* working the same as ttrpg. the direction for medical isn't entirely set if we want to do conditions or hp yet (basically-should we track health via literal numbers or through symptoms, wounds, etc), so im kinda wishy washy on the concept rn. i at least do want this number to be a little obscured, i dont really want players to just be able to say "im at 8 vitality right now so i can get another upgrade." the number should be obscured, with extra factors that impact it, so there's at least a little back and forth

Terkala:
@taydeo before I put a bunch more time into this, is this accepted as a design PR for medicine/surgery going forward?

Terkala:
Skills "entirely" removed. All surgery is purely based on tools. Roboticists will just get a special tool, and doctors will have their surgery rooms provide the bonuses doctors used to get.

taydeo (approved):

PR #44, opened 2026-01-01
Reagent Rework

Homingpenguins:
The Reagent Rework Added to Design Proposals on request of Mkanke.


corresp0nd (commented):
i dont disagree with any of this, cause yeah. it should all just work this way already. my main concern when people bring up this discuss is well, how are you going to fix it though? reagent code is kind of a mess as it is and personally needs to be reexamined from the ground up. i think im more impartial to just seeing what direction nubody takes reagent code and then considering if that suits our needs. it might not, but im pretty sure the lava beaker stuff is at least pretty high up there as a concern. maybe. who knows.

PR #45, opened 2026-01-06
reagent piping

Chedd-Error:
A design doc i spent a morning on to conceptualize down how reagent pipes would work for funky. Goes over the bare minimum of the system and some basic interactions that I can think of for it.

### DIFF
--- a/src/design-proposals/reagentpiping.md
+++ b/src/design-proposals/reagentpiping.md
@@ -1,0 +1,162 @@
+# Reagent Pipes
+
+| Designers | Implemented | GitHub Links |
+|---|---|---|
+| zergologist | :x: No | TBD |
+
+## Overview
+
+A reagent pipe system quite similar to the existing atmos pipe system, utilizing the same pipe layers and allowing for reagents to be pumped throughout the station.
+
+## Background
+
+Given that, on a space station in the middle of the void, there being an infinite and free source of water through the sinks seems very gamey. Drains just removing a set amount of liquid reagents permanently doesn't help with this feeling. The fact that the only variety of plumbing is for gas pipes is also strange. Given that atmos pipes were recently updated to allow for three layers of piping or a total of up to six individual pipe networks crossing one tile, adding in a new system that utilizes the same pipe layers to make actual plumbing work wouldn't interfere with the station's atmos to a severe degree.
+
+Making atmos pipes look less like actual pipes and more like ductwork in order to differentiate the two nets would likely be needed.
+
+## Features to be added
+
+### Reagents in Pipes
+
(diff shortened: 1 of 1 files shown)


corresp0nd (changes requested):
ive considered something before and i personally came to the conclusion that i dont think such a system is particularly worth it. particularly, i dont know what a system to add an entire pipenet would add when this would largely just end up being dlc for the chemist to play with. a lot of this is going to culminate into just letting the chemist store and produce a lot of reagents at once, which isnt a bad thing, really. i think that could be fun, especially if we decided to do some sort of reagent bounty. but making it into a pipenet that everyone is affected by brings up a lot of concerns that i don't think make this addition, as-is, worth it. 

1. how would we implement entire networks of pipes without it being too much of a mapping burden?

consider train. people dont like playing on train and it isnt in rotation for a reason, but look at the ways the mapper tried to solve the space problems and how much content had to be cut just because of space problems. pipe layers make reagent piping feasible, yes, but there's still a pretty large overhead for something that doesn't add a lot in the first place. 

2. what does this even add? 

you proposed a new plumber role, but like. what is a plumber going to do. turn on the big valve that says "water" and walk away? they could be the ones tasked with repairing it, sure, but in an emergency situation where there's water leaking everywhere then that's just going to be done by whoever can fix it. this doesn't need its own role.

waste lines aren't going to give a plumber much to do, either. like, honestly, i dont think a waste line needs to exist at all. i agree drains should be reworked (or at least refactored because they are mini lag machines) but i dont think that needs to be done in this way. the only person who ever has waste chemicals are the chemist or the bartender, both of which should be discouraged from wasting things in the first place. they don't need a way to get rid of excess chemicals (unless something has gone wrong, like duplicating chemicals). 

3. why does it even have to be pipes?

like. okay. think of it this way. we're on a space station in space. ran by a megacorp. do you think nanotrasen gives a fuck about making sure every crew has access to drinkable water. i personally don't, at least not to the point where they're installing very complex water systems in a very dangerous environment. i think they're giving everyone barrels of filtered water and calling it a day.

so i dont think it Needs to be literal pipes. i dont think sinks or drains have to really exist outside of being a suspension of disbelief. like, bathrooms exist so there's gonna be a way to wash your hands. in a sink. that does not to be more complex than it already is because no one washes their hands in a video game where you physically cannot press the button to wash your hands. and thats fine because we dont even have a reason for players to wash their hands. my point is like. why are we trying to make this very complex system so people can use a sink when there's no point in using a sink in the first place. 

4. why do we need to rebalance water?

having infinite ammounts of water is "abusable" in like. two ways. we do not need to rebalance how people get water because it's fucking water. nothing is going to change if we make it more or less hard to get water. Oh no the guzzler is going to drink 1000u of water and ruin the supply system!!! What. Who cares. I do not care if the guzzler is guzzling water.

PR #46, opened 2026-01-07
Holy Damage Redesign

SigmaTheDragon:
Holy damage is the most useless damage type of them all. Please, i'd rather deal ion than holy damage ;-;

Will also serve to rebalance the chap powercreep so that they deal more with the antags they are supposed to deal with.


corresp0nd (changes requested):
i dont disagree that chaplain should have special interactions with the more magical antags, but im worried about creating this whole entire System just for that one interaction. like, i dunno. i feel like this would just be better if we were trying to make a game about magic, and i dont really know if that's we want in the first place. implementing a system that's probably going to be really annoying to balance around for something that we would probably consider a side aspect of the story is kind of a tall ask, personally

PR #47, opened 2026-01-12
Create syringe-dirtiness

Queer-Cats:
design doc for syringe dirtiness

### DIFF
--- a/src/design-proposals/syringe-dirtiness
+++ b/src/design-proposals/syringe-dirtiness
@@ -1,0 +1,47 @@
+# Syringe Dirtiness
+
+| Designers | Implemented | GitHub Links |
+|---|---|---|
+| Queer Cats | :x: No | TBD |
+
+## Overview
+
+Adding dirtiness as a mechanic to syringes. Syringes get dirtied upon use, and deal poison damage to patient when injected while dirty. 
+
+## Background
+
+Syringes currently dominate the med treatment meta, overshadowing most other methods of treatment, such as pills, autoinjectors, IV bags, surgery, cryogenics, etc. This is the result of syringes being very fast acting, effectively free, and with no major downsides. This is further compounded by there being no downside to drawing from large containers, so medical tends to just keep jugs or even buckets of chems in an open area and having all doctors draw from the same pool, which would be a sanitation nightmare if done irl/.
+
+We can take some inspirations for nerfs from the real world, where syringes are typically avoided as a delivery method where possible because they are invasive and prone to causing infection when not properly handled. The former is not easily modeled given the current medical system, so the focus here will be on the latter.
+
+## Features to be added
+
+The primary feature addition would be a cleanliness mechanic similar to that currently in use for surgical tools, with some modifications to reflect syringes having less individual actions. 
+
+All effects should occur after the relevant DoAfter so both the doctor and the patient have a chance to avoid the effcts. 
+
+Syringes would deal poison damage to a patient after injecting or drawing from them, depending on the dirt level. At very low levels, there is no poison, above a certain threshold, damage scales linearly, with max dirtiness doing 200 poison damage, requiring surgery to correct. 
+
+Syringes should immediately max out dirtiness after being used on a living thing, and after drawing from a puddle. An alternative proposal for puddles is to have any chemicals in puddles slowly convert to mould over time as an alternative way to punish injecting medicines directly from the floor into a patient.
+
+Syringes should gain some dirtiness when drawing from a container, scaled to container size to encourage usage of smaller containers. It should be possible to draw from a vial or bottle up to three times without hitting the threshold for poisoning to allow doctors to tailor doses appropriately. Drawing from a bucket should immediately max out dirtiness. 
+
+Hyposprays and autoinjectors would be sterile, so as to allow nukies to continue to use jugs for space effeciency, and to give autoinjectors a greater usecase.
+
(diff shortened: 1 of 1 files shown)


corresp0nd (changes requested):
some considerations that need to be addressed:
- how easily accessible are syringes going to be? like are there going to be packs of them that are all disposed, or are we swapping out the needles after every use...
- how are we going to dispose syringes and handle contaminated materials in general?
- are syringes going to use the same dirtiness system? if they are i want us to reexamine the system and what it intends to do in the first place
- what are the alternative ways to deliver chemicals? should we reconsider those to make them more interesting and fun to use rather than just trying to make a certain option worse? 

i dislike how syringes are currently used, but i also recognize that like. its kind of a bad system overall, not just specifically the syringes. i want us to reconsider the interaction as a whole (administering medicine) and figure out what parts we genuinely like or dislike, both tonally and gameplay wise

src/design-proposals/syringe-dirtiness
```diff
@@ -0,0 +1,47 @@
...
+
+All effects should occur after the relevant DoAfter so both the doctor and the patient have a chance to avoid the effcts. 
+
+Syringes would deal poison damage to a patient after injecting or drawing from them, depending on the dirt level. At very low levels, there is no poison, above a certain threshold, damage scales linearly, with max dirtiness doing 200 poison damage, requiring surgery to correct. 
+
+Syringes should immediately max out dirtiness after being used on a living thing, and after drawing from a puddle. An alternative proposal for puddles is to have any chemicals in puddles slowly convert to mould over time as an alternative way to punish injecting medicines directly from the floor into a patient.
+
+Syringes should gain some dirtiness when drawing from a container, scaled to container size to encourage usage of smaller containers. It should be possible to draw from a vial or bottle up to three times without hitting the threshold for poisoning to allow doctors to tailor doses appropriately. Drawing from a bucket should immediately max out dirtiness.
```
corresp0nd, line 27: a perfectly sterile syringe drawing from a container that has either never been drawn from before or has only been drawn from other sterile equipment should not cause dirtiness. consider letting objects transfer dirtiness from each other instead, and maybe like. letting vials/bottles be "capped" so they can only be drawn from (think of actual vials you draw medicine from, they have a little membrane over them that doesn't let liquid through while still letting you poke it with a needle)

src/design-proposals/syringe-dirtiness
```diff
@@ -0,0 +1,47 @@
...
+
+We can take some inspirations for nerfs from the real world, where syringes are typically avoided as a delivery method where possible because they are invasive and prone to causing infection when not properly handled. The former is not easily modeled given the current medical system, so the focus here will be on the latter.
+
+## Features to be added
+
+The primary feature addition would be a cleanliness mechanic similar to that currently in use for surgical tools, with some modifications to reflect syringes having less individual actions. 
+
+All effects should occur after the relevant DoAfter so both the doctor and the patient have a chance to avoid the effcts. 
+
+Syringes would deal poison damage to a patient after injecting or drawing from them, depending on the dirt level. At very low levels, there is no poison, above a certain threshold, damage scales linearly, with max dirtiness doing 200 poison damage, requiring surgery to correct.
```
jackel234: i think personally, until we have proper systems in place to represent infections, syringes should just become unusable after they are used

jackel234:
> * how easily accessible are syringes going to be? like are there going to be packs of them that are all disposed, or are we swapping out the needles after every use...

i'm not the original author, but i personally have had some similar ideas on this, specifically making syringes a disposable item. i think that syringes should come in sterile packages, which can either be gotten from the nanomed, or printed from the medfab. i think they should leave trash behind once they are unwrapped, because i think the visual of a post-rush med being covered in trash from sterile packaging and blood would be extremely interesting :) 
additionally, i think syringes should be single use, just like real life, excluding possibly advanced syringes made by science. i'm of the opinion that we could also have small syringe pouches that hold anywhere from like 5-15 syringes at once and can be put in your pocket, that or have a spot on a crash cart for holding syringes, just to make accessing them a bit easier. 

> * how are we going to dispose syringes and handle contaminated materials in general?

sharps containers! and specific biohazard trash cans for other biological materials, like used gloves. my idea for sharps containers is that they can be sold once they are full, so you can get back something from disposing of them properly. but yeah, when i think of a hospital, or really any medical facility, rooms with the bright red sharps containers are what first come to mind for me

Queer-Cats:
> some considerations that need to be addressed:
> 
>     * how easily accessible are syringes going to be? like are there going to be packs of them that are all disposed, or are we swapping out the needles after every use...
> 
>     * how are we going to dispose syringes and handle contaminated materials in general?
> 
>     * are syringes going to use the same dirtiness system? if they are i want us to reexamine the system and what it intends to do in the first place
> 
>     * what are the alternative ways to deliver chemicals? should we reconsider those to make them more interesting and fun to use rather than just trying to make a certain option worse?
> 
> 
> i dislike how syringes are currently used, but i also recognize that like. its kind of a bad system overall, not just specifically the syringes. i want us to reconsider the interaction as a whole (administering medicine) and figure out what parts we genuinely like or dislike, both tonally and gameplay wise

Yeah, I've sort of put this idea on hold until FunkyMed becomes more developed because I think this is by itself a somewhat band-aid solution onto a fundamentally flawed system. It's possible that with other changes to the medical system, this becomes less necessary.

PR #48, opened 2026-01-13
conspirators

Chedd-Error:
my own draft of how conspirators for funky should be implemented, contains quite a few sample objectives that can be sorted through and altered in case any don't fit funky's direction as they currently are

### DIFF
--- a/src/design-proposals/conspirators.md
+++ b/src/design-proposals/conspirators.md
@@ -1,0 +1,54 @@
+# Conspirators
+
+| Designers | Implemented | GitHub Links |
+|---|---|---|
+| zergologist | :x: | TBD |
+
+## Overview
+
+Adds in [Conspirators](https://wiki.ss13.co/Conspirator) to Funky, roleplay focused team antagonists ported from Goonstation 13. They are intended to bring a more RP-focused conflict into the shifts that they're present, either by infiltrating specific departments, attempting to convince the station of one inane thing, or by otherwise impacting the overall flow of the round in a silly yet disruptful way.
+
+## Background
+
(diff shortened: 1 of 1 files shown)


corresp0nd (changes requested):
this is almost-approved but i have some tentative questions and concerns. namely, im iffy on it being a literal antag when it doesn't seem like it does very many antag things. the concerns you listed, about people critiquing the original implementation on goon, are true, and i don't think a different culture is just going to magically fix those concerns. i would like to see more thought about how we could possibly address those without taking away the base idea of the concept. which, as im understanding it, is just a crewmember who has a specific goal that is probably against nt's and command's wishes, but not necessarily an actual harm to the station. 

i would also like to have an admin review this and see if there are any concerns about how freeform the antag is. we have had issues with this before and i don't want those same issues to be repeated. @PoquitoDondito @salpphie

salpphie:
My concerns/thoughts are as follows:

1. As a team antagonist, do they have the ability to talk to each other, or are they like tots who have to figure it out via codes? This would impact them as a team antagonist if they don't have the means to communicate with one another easily, or having too different operations from one another. 

2. To what level can the conspirators engage in violence, having such free-form ideas can lead to people using said objectives and stretching it to meet said objectives. Should they be viable to engage in round removal, given they can do a life-or-death gameshow, or is it tot RR rules where you can only engage in RR behavior if you have the objective specifically?

3. Having a bunch of them seems concerning if they're going to spawn alongside other antags like tots, changelings, heretics, etc. Maybe wrangle it down to like. 3 max.

4. Why are they sowing seeds of chaos? Are they with a Syndicate faction? Do they have a major reason to be doing such things.

I probably have more, but these are my thoughts/concerns for now from an admin perspective. Already thinking about how this will appear in a guidebook and such.

PR #49, opened 2026-01-14
Condition-based medical

otokonoko-dev:
Please feel free to discuss anything and ask for clarification.

### DIFF
--- a/src/design-proposals/condition-based-medical-system.md
+++ b/src/design-proposals/condition-based-medical-system.md
@@ -1,6 +1,6 @@
 # Condition-Based Medical System
 
-| Designers | Implemented                    | GitHub Links |
+| Designers | Implemented | GitHub Links |
 | --- | --- | --- |
 | Ven | :x: No | TBD |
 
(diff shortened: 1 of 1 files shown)


miosar:
I have 3 questions for this:
Would our current chem system be changed? I.E, salic acid treats brute damage. Would it mend broken bones and open wounds without surgery or would it act as a "minimal" temporary treatment?

Will the condition system be balanced around self sustainability? This is mainly based off of offmed concerns, where most antagonists and security officers often died because of conditions they could not treat in the field due to being considered a station wide threat and/or limited supplies ( such as heretic only having omnizine from the flask ritual and whatever they steal from medical )

Mostly concerned about you mentioning that permament solution are limited to medical. While I support this motion in the future of funkymed, it is important to remember that access to professional medical help is very often limited, even to a permament degree. A prisoner will be kept in the brig in cuffs, nukies are forced to remain on the move, heretics and syndies going loud are limited by TC or their abilities and medical itself is already often overwhelmed.

In short, this line:

`- Players should not be able to fully diagnose or optimize their treatment without medical involvement.`

Worries me, due to both the experience and ability of medical players (both new and old) to deal with station threats and them being (understandably so) not supposed to help known threats to both others and themselves (such as, again, heretics, loud syndies, nukies, and in the future revolting unions of any department outside of medical. 

Oh, also, small question at the end: What will prevent players from just figuring out their damage using just simple logic? As in, "I got hit with a toolbox in the head, my vision is blurry, therefore I probably have a concussion." or "I got shot and have trouble breathing. I probably have a punctured lung."

Terkala:
The one thing I notice as being suspiciously absent from this proposal: What does "crit" or "dead" look like under this system? How does it determine a player has gone into crit or died? Can you break every bone in a person's body and have them not die? How does it work when you take a type of damage that normally "would" break a bone, but the bone is already broken?

I know it's listed under open questions, but this is pretty core to even being able to start on something like this

otokonoko-dev:
> The one thing I notice as being suspiciously absent from this proposal: What does "crit" or "dead" look like under this system? How does it determine a player has gone into crit or died? Can you break every bone in a person's body and have them not die? How does it work when you take a type of damage that normally "would" break a bone, but the bone is already broken?
> 
> I know it's listed under open questions, but this is pretty core to even being able to start on something like this

Good feedback. I did have a vision on how death/crit works and I don't quite get how i missed adding it to the doc. I've updated the document to add a section about how i envision crit/death to be handled now though.

corresp0nd (approved):
this is a great first start and i think it pretty well explains the core idea that we've been trying to get across. the concerns mentioned prior are still true, so i would like to see those worked on at some point. but obvi this is just one step of many so like, whatever...
one thing i would like to see discussed next is how we handle player agency as someone without access to medical tools, particularly regarding pain. how to we communicate pain to players without it being obtrusive and feeling forced, while also not letting it be easy to forget about? the pain system is a weird system where we still need to respect the fact that the player is not their character, while still communicating that a character has to feel this specific thing. its important that we properly think about this before moving on

PR #52, opened 2026-01-21
Cyborg Additions Proposal

TheHolyAegis:
I have the feeling that borgs are a little monotone when you have played them for a while. So I have come up with some changes to hopefully improve the variety you can play them as.

I used the design doc that was already made in #19, and changed it up a little to update it to the changes it has gotten already, and some new ideas I came up with.

More ideas and suggestions are always welcome.

### DIFF
--- a/src/design-proposals/cyborg-redesign.md
+++ b/src/design-proposals/cyborg-redesign.md
@@ -23,7 +23,7 @@ The summary of the changes are as follows:
 ***
 ### Engineering Chassis
 
-This chassis already seems perfectly balanced, I want the other chassis need to get buffs to bring those chassis to the same level as this chassis in terms of usage.
+This chassis already seems perfectly balanced; I want the other chassis to get buffs to bring those chassis to the same level as this chassis in terms of usage.
 
 ***
 ### Salvage Chassis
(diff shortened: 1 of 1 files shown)


corresp0nd (approved):
i personally think that this isn't much of a rework and more so of just like. a tweak to how they currently work. that's fine. this is all fine. you gave them more toys and that's okay 

if you wanna really rework them i would consider what role they currently fulfill in the game (its kind of a bad one honestly, borgs are pretty weak imo) and how they could be better

TheHolyAegis:
I mostly just wanted to give them more toys. I noticed ever chassis always ends up with the exact same modules, and they almost never reach the module limit. They (or the roboticist) never have to choose which modules they want. I feel like expanding the amount of modules would make borgs more diverse in their usage. I also feel like more roundstart/latejoin cyborg slots would be nice, because currently maps have at most two or three built-in. Though that isn't strictly necessary.

I would still be thinking what modules I would add for every chassis. Maybe a whole new chassis for maybe security? Maybe I would add more modules that can be added to any chassis. But I will try to improve the borg play experience.

TheHolyAegis:
I think I'll wait with implementing this until after forky station, because then I think whitelisted hands exist in borgs? I sure hope so.

PR #53, opened 2026-01-22
Central Command Revamp Proposal

TheHolyAegis:
I saw in a pull request people talk about a change to Central Command. I thought to make a proposal so the conversation can be moved to here.

I made the doc with the jist I got from the conversation, and added some of my own thoughts in it as well.

(The PR where the earlier discussion took place: https://github.com/funky-station/funky-station/pull/2714)

### DIFF
--- a/src/design-proposals/cc-revamp.md
+++ b/src/design-proposals/cc-revamp.md
@@ -1,0 +1,65 @@
+# Short, Properly Capitalized Title
+
+Your title should convey the basic jist of your proposed changes. It should be short because the text will be linked in the sidebar.
+
+| Designers | Implemented | GitHub Links |
+|---|---|---|
+| TheHolyAegis | :x: | TBD |
+
+## Overview
+
+This proposal is about the idea to make Central Command like the ghost bar, or at least make it a space for dead people (or observers) to chat around and do something while they wait for the round to end.
+
+## Background
+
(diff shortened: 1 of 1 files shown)


corresp0nd (changes requested):
the ghost bar does suck and no one uses it because it sucks. i would personally like to see more thought on why the ghost bar fails in this first place, because i personally do not feel like this really addresses Why something like the ghost bar sucks. making it into CC sort of just slaps a new coat of paint onto it. 

also, when we make decisions like this regarding central command we also have to understand that blurring the line between player and admin is not always a good thing. yes, the game should be immersive so you can play out your stories, but admins are OOC because they usually have to talk to you about important things that need to be addressed OOC. if you get a fax from CC and you aren't sure if it's another player or admin, but it says "disregard all orders and nuke the station" then that's an issue. you should know right away if this is from an admin or not and how seriously you should take that fax. the same with getting at CC, post round, and seeing a CCO saying something important. there should be clear lines between when something is important, and the distinction between central command and the station has ALWAYS been the way to draw that line. we do not need to blur it.

TheHolyAegis:
Yeah, true. And I have already helped a little with that. The CCI now has their own icon and stamp, so it's more obvious to the station crew when it's an admin, or another player.

PR #55, opened 2026-01-25
Species design document: Martians/Eggheads

Teasq:
https://www.youtube.com/watch?v=yX-SdAj_oJo


corresp0nd (approved):
yup theyre cute. APPROVED!

regarding the language concerns: eh. i dont really think we need an entire language system for them. the regex for accents is REALLY strong, you can just make a fucking massive one (and maybe utilize the markup tags to change the chat font) to make it seem like it's an actual language. this is what greys from impstation do and its always been my favorite part of them. i also think its cute if its like 90% gibberish but you can understand these few words, fun for players to try and decode just a tiny bit of what theyre saying, too.

PR #57, opened 2026-01-26
IC Guidebook

mkanke-real:
First draft of a guideline for future guidebook development


corresp0nd (approved):
ive really wanted this to be a big goal of ours so im completely on board. the only thing is that i do want there to be an extra system to make it more accessible to new players rather than just "remember everything." like the information should still be available in game somehow, even if you cant get access to the book. maybe just SOPs? i dunno. they should still be formatted IC and all i just dont want people to be forced to memorize everything 

also your other design doc is here so that should be fixed. blame git

mkanke-real:
> ive really wanted this to be a big goal of ours so im completely on board. the only thing is that i do want there to be an extra system to make it more accessible to new players rather than just "remember everything." like the information should still be available in game somehow, even if you cant get access to the book. maybe just SOPs? i dunno. they should still be formatted IC and all i just dont want people to be forced to memorize everything
> 
> also your other design doc is here so that should be fixed. blame git


I could definitely see the need for that. I would say a good middle ground is ensuring every locker comes with the associated guidebook and I could get the mapping team to add some hardmapped books for jobs that lock a proper locker. Would you say that would be sufficient to ensure that new players can find their info? 

I really just want to nuke the guidebook button as much as possible with IC info in IC books so people understand everything in that tab is OOC info and should not be talked about in things like help chat or in game.

The other file should be removed now too btw

PR #58, opened 2026-01-27
Corporate Interests proposal

dudeheadface:
design doc to give NTRs "corporate interest" objectives

### DIFF
--- a/src/design-proposals/corporate-interests.md
+++ b/src/design-proposals/corporate-interests.md
@@ -47,6 +47,8 @@ Loyalty testing: Orchestrate the conducting of thorough loyalty tests for all co
 
 Make an example: You are to find a fitting candidate for execution by the end of the shift. Take your time and ensure this is not done baselessly but some bending of the usual requirements will be approved. This is to be approved and overseen by the magistrate if available. If no magistrate is available you must fax CC to approve your candidate. This execution is to be done publicly with an announcement of their crimes against NT. These crimes may be embellished to avoid an uproar but the message regarding acting out against NT should be clear.
 
+Secure and contain: We believe there may be dangerous individuals on board with knowledge or abilities that may prove to be valuable assets in our research of the unknown. At least one of these individuals must be contained and brought alive to the CC outpost to be transported to a secret research facility. The abilities they possess may not be removed or suppressed. (Can appear in rounds with changelings, wizards, heretics, or either forms of cultists).
+
 Interests like “drug testing” may require additional features being added. Features like this could very well see use outside of the interest they were created for. Drug testing’s randomized chemical potentially being found in maints pills for one example.
 
 Planned future content will allow for more and more memorable interests to be added. Xenobiology or virology forced to create bio weapons for NT, genetics making super soldiers or mutations that suppress free will in some way, payment of scribs could be docked from the department deemed least productive and given to that deemed most productive in an “incentivise competition” interest etc.
--- a/src/design-proposals/corporate-interests.md
+++ b/src/design-proposals/corporate-interests.md
@@ -37,7 +37,7 @@ Here are some examples of corporate interests:
 
 Lost command manual: It has been reported that a lost command manual containing corporate secrets is somewhere on board. Secure it and silence any crew deemed a liability to maintaining these secrets. Do this by any means necessary. Take no risks.
 
-Zombie specimen: A small amount of the bio weapon drug known as romeral has been found on station and is now in your possession. Ensure that a sapient being on board is infected and their condition progresses to the final stage of zombification. Ensure they are secure and do not allow this infection to spread. They must be discreetly brought in “live” zombie form to the CC outpost so they can be transported to a specialized research facility.
+Zombie specimen: A small amount of the bio weapon drug known as romeral has been found on station and is now in your possession. Ensure that an animal on board is infected and their condition progresses to the final stage of zombification. This condition must progress through each stage of infection so ensure the animal is not killed through any other means. This animal must be larger than a mothroach and may not be cube grown. Ensure they are secure and do not allow this infection to spread. They must be discreetly brought in “live” zombie form to the CC outpost so they can be transported to a specialized research facility.
 
 Drug testing: The following chemical drug is in the experimental stage and we need results from human testing (drug is a chemical with randomly generated effects). Divulging this to as few members of the crew as possible, ensure this drug is discreetly tested on X amount of people with the results recorded and faxed to CC.
 
@@ -45,9 +45,11 @@ Downsizing: Ensure that X amount of crew are fired before shift’s end includin
 
 Loyalty testing: Orchestrate the conducting of thorough loyalty tests for all command members and X amount of crew. You are not required to conduct these tests personally. Just see that it is done. Documentation of each test including a loyalty rating from 1 to 5 is to be faxed to CC.
 
-Make an example: You are to find a fitting candidate for execution by the end of the shift. Take your time and ensure this is not done baselessly but some bending of the usual requirements will be approved. This is to be approved and overseen by the magistrate if available. If no magistrate is available you must fax CC to approve your candidate. This execution is to be done publicly with an announcement of their crimes against NT. These crimes may be embellished to avoid an uproar but the message regarding acting out against NT should be clear.
-
 Secure and contain: We believe there may be dangerous individuals on board with knowledge or abilities that may prove to be valuable assets in our research of the unknown. At least one of these individuals must be contained and brought alive to the CC outpost to be transported to a secret research facility. The abilities they possess may not be removed or suppressed. (Can appear in rounds with changelings, wizards, heretics, or either forms of cultists).
+
+Major deficit: Our regional commander accidentally misplaced 200000 spesos during the previous quarter and we need a windfall to make up that deficit. You are to obtain that money and arrive with it in your briefcase by any means necessary.
+
+Priviliged hospitality: The regional commander had to leave the sector on business and their son is bored. You are to protect and entertain his son by any means necessary" (Spawns a ghost role for the son with the objective of "be as shitty and demanding of the station as possible")
 
 Interests like “drug testing” may require additional features being added. Features like this could very well see use outside of the interest they were created for. Drug testing’s randomized chemical potentially being found in maints pills for one example.
 
--- a/src/design-proposals/corporate-interests.md
+++ b/src/design-proposals/corporate-interests.md
@@ -49,7 +49,7 @@ Secure and contain: We believe there may be dangerous individuals on board with 
 
 Major deficit: Our regional commander accidentally misplaced 200000 spesos during the previous quarter and we need a windfall to make up that deficit. You are to obtain that money and arrive with it in your briefcase by any means necessary.
 
-Priviliged hospitality: The regional commander had to leave the sector on business and their son is bored. You are to protect and entertain his son by any means necessary" (Spawns a ghost role for the son with the objective of "be as shitty and demanding of the station as possible")
+Priviliged hospitality: The regional commander had to leave the sector on business and their son is bored. You are to protect and entertain his son by any means necessary. (Spawns a ghost role for the son with the objective of "be as shitty and demanding of the station as possible")
 
 Interests like “drug testing” may require additional features being added. Features like this could very well see use outside of the interest they were created for. Drug testing’s randomized chemical potentially being found in maints pills for one example.
 
--- a/src/design-proposals/corporate-interests.md
+++ b/src/design-proposals/corporate-interests.md
@@ -1,0 +1,95 @@
+# Corporate Interests
+
+
+| Designers | Implemented | GitHub Links |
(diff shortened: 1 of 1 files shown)


Gorsthir:
I do not think giving job roles objectives is the way to go here. My main issues with this:

1. Put too much focus on one specific job. Promotes "main characters", which is something we typically want to avoid.
2. Having specific coded objectives per round is limiting and having them for a job role with a lot of authority makes it too much of a driving force in directing where the round goes. I find this limits both individual creativity and autonomy, as well as spontaneous roundflow. The direction the round takes should in my opinion be driven by collaborative storytelling where characters naturally respond to the actions of others, not mechanically enforced.
3. Some of the examples given are things that would typically be admin driven events, which there are guidelines around to ensure it is properly planned and carried out with admin oversight.

All this to say, I love people being creative and wanting to do something that deviates from the regular roundflow every now and then. I am however strongly against mechanically enforcing it like this. Let it instead just be player driven on their own initiative, with people ahelping to ask if they need permission to do something a bit out of the norm.

A-Loose-Goose:
I actually quite like this par one thing
Romerol the the NTR
Dare I say how this can go bad and completely change a round and ruin certain things? ONTO A CREW ALIGNED MEMBER? Not only can this kill and ruin many antag rounds but could also cause Zed ops to just fail due to zombies being on station already
Think that needs to go or be reimagined

dudeheadface:
> I do not think giving job roles objectives is the way to go here. My main issues with this:
> 
> 1. Put too much focus on one specific job. Promotes "main characters", which is something we typically want to avoid.
> 2. Having specific coded objectives per round is limiting and having them for a job role with a lot of authority makes it too much of a driving force in directing where the round goes. I find this limits both individual creativity and autonomy, as well as spontaneous roundflow. The direction the round takes should in my opinion be driven by collaborative storytelling where characters naturally respond to the actions of others, not mechanically enforced.
> 3. Some of the examples given are things that would typically be admin driven events, which there are guidelines around to ensure it is properly planned and carried out with admin oversight.
> 
> All this to say, I love people being creative and wanting to do something that deviates from the regular roundflow every now and then. I am however strongly against mechanically enforcing it like this. Let it instead just be player driven on their own initiative, with people ahelping to ask if they need permission to do something a bit out of the norm.


I get where you’re coming from but I do disagree.

1. I don’t believe this would put as much focus on the NTR as you’re implying. Yes, most of the examples I gave were more extreme but the majority of objectives would be on the more mundane side and in either case they are meant to be fulfilled more or less in the background of the round as a whole (unless something goes wrong but that’s interesting dynamic round flow in my opinion). I don’t really get how this would cause any main character syndrome even in the more extreme objectives. It’s not about NTR having cool stuff to do it’s about creating interesting scenarios for all involved.
2. I believe it would be the opposite of limiting. It opens up new avenues for roleplay and character interaction. The conflicts that would naturally arise encourage collaborative story telling with the objectives as a catalyst, not restriction. Every round’s story is already mechanically driven by antags and the roles of the jobs around the station. Mechanics give a level of structure and unpredictability that creates that dynamic storytelling environment.
3. Yes, this is why I tried to be extra careful in my wording of those objectives. I get where the worry comes from but the potential benefit of making it work I think is worth the effort trying to get there.

I believe these objectives encourage creativity and dynamic player interaction. It’s the same way that antags are designed to add excitement and variety to the round. I believe the hand of NT should act in a similar way. 
The vast majority of players won’t ahelp for things like this and as you said even when they do it’s usually only for things a BIT outside of the norm. Many of the more exciting and interesting potential objectives will never happen without some level of mechanical enforcement

dudeheadface:
> I actually quite like this par one thing Romerol the the NTR Dare I say how this can go bad and completely change a round and ruin certain things? ONTO A CREW ALIGNED MEMBER? Not only can this kill and ruin many antag rounds but could also cause Zed ops to just fail due to zombies being on station already Think that needs to go or be reimagined

I actually think this would be an extremely rare scenario. The objective would likely be very rare, theres already many things in game that are way more likely to “ruin” antag rounds including other antags in that round, there is an equal chance that even in the case of a full outbreak alongside other antags this would add to the round by introducing an unexpected factor to the situation. Let’s not forget there’s a tiny chance for romeral to be inside maints pills.

On taking issue with this being in the hands of a crew aligned player. This proposal comes from the perspective that NT really shouldn’t be seen as crew aligned. They're not antags but they should have conflicting interests with that of the crew. There was a great discussion I think in the discord devchat about how it might be a good idea to separate the alignments that are now considered crew into crew/station/NT with station being sec and command and NT being CC.

I may very well be wrong and this objective would cause too many issues but I think it’s worth testing at least.

Idem778:
> "The regional commander had to leave the sector on business and their son is bored. You are to protect and entertain his son by any means necessary" (This one would spawn a ghost role of the son with the objective of "be as shitty and demanding to thw crew as possible" or something. This idea was stolen whole sale from an admin event thought up by MK :godo:)

Yes, yes yes yes, all of the yes. This would be so funny

corresp0nd (changes requested):
i agree with the general concept, but i do think that something like this needs to be considered slightly more than just giving the ntr their own objectives. my major concern is that this should really be its own major gamemode, not really just something that can happen every shift. 

this is primarily because all of your objectives are really round influencing and need their own guidelines and exclusions to be done well, they cant just be part of one big thing. dont get me wrong it could Theoretically cause some fun situations, but like. we can let this be its own story. this also gives you a little more leeway to let it be a group objective, rather than just ntr. magi and cappy, maybe iaa, should be part of this, too, rather than just the ntr (which should also hopefully address some of the main character concerns).

dudeheadface:
> i agree with the general concept, but i do think that something like this needs to be considered slightly more than just giving the ntr their own objectives. my major concern is that this should really be its own major gamemode, not really just something that can happen every shift. 
> 
> this is primarily because all of your objectives are really round influencing and need their own guidelines and exclusions to be done well, they cant just be part of one big thing. dont get me wrong it could Theoretically cause some fun situations, but like. we can let this be its own story. this also gives you a little more leeway to let it be a group objective, rather than just ntr. magi and cappy, maybe iaa, should be part of this, too, rather than just the ntr (which should also hopefully address some of the main character concerns). 

I don't think I fully understand the issues you're bringing up. I did say in the doc that although the majority of examples given were higher round impact
objectives, most actual objectives would be far lower impact like "loyalty testing". I also feel that even the more extreme objectives aren'l likely enough to have the consistent level of impact to warrant being it's own game mode. 

The idea behind many of the objectives is supposed to be that the interests of NT would often make a bad situation worse. Think the one corporate rep in aliens trying to get Ridley and Newt implanted with a xeno after shit already hit the fan. I think the idea of the crew at large being stuck between the threat of antags and the uncaring arm of NT is where this proposal brings the most potential. 

As for it not being a full group objective, is the issue that the NTR would technically be the only one that would succeed or fail? Because in practice I thought it would already basically work like a team objective as the only real way for the NTR to complete many of them is by leveraging their position to have IAAs, members of command, or potentially even a few crew members do the actual hands on work. It's sort of a dynamic team objective lead by the NTR as apposed to the exact same roles involved every time which I think is more limiting to potential role play and conflict.

Sorry, I don't mean to be rejecting of criticism I just really want to grasp what the issues are and I don't think I am. Also if every round is too much it could be something that's rolled sort of like a minor antag or even just admin enabled. Though, personally I would advocate for a trial run incorporating it into regular gameplay to see if it really is an issue.

corresp0nd:
> I don't think I fully understand the issues you're bringing up. I did say in the doc that although the majority of examples given were higher round impact objectives, most actual objectives would be far lower impact like "loyalty testing". I also feel that even the more extreme objectives aren'l likely enough to have the consistent level of impact to warrant being it's own game mode.

since there are high impact objectives that will need their own rulings and considerations, then we should just treat it like its own game mode. like, think of it this way, DAGD can drastically alter a round, but the fact that it's a lot rarer doesn't mean we don't consider it apart of the traitor gamemode. it doesn't matter that DAGD/high impact objectives are rarer, but we still consider all of them as part of the same gamemode. if we really want to keep something like this as an every round thing, then i think we're going to have to be really strict with what we consider "low impact objectives." like for example i would personally not include the ceo's son objective in the low impact pool, even though it doesn't really Do a lot. 


> The idea behind many of the objectives is supposed to be that the interests of NT would often make a bad situation worse. Think the one corporate rep in aliens trying to get Ridley and Newt implanted with a xeno after shit already hit the fan. I think the idea of the crew at large being stuck between the threat of antags and the uncaring arm of NT is where this proposal brings the most potential.

yeah, i agree! i think this is fun and cute. what i'm trying to say is that this shouldn't happen every round. this can be a really common gamerule, that's fine! 

> As for it not being a full group objective, is the issue that the NTR would technically be the only one that would succeed or fail? Because in practice I thought it would already basically work like a team objective as the only real way for the NTR to complete many of them is by leveraging their position to have IAAs, members of command, or potentially even a few crew members do the actual hands on work. It's sort of a dynamic team objective lead by the NTR as apposed to the exact same roles involved every time which I think is more limiting to potential role play and conflict.

if it in practice works a specific way, we should codify that to avoid player confusion. if we don't then there is always the chance that someone is going assume "well, im a magi and i only care about space law. fuck nt's interests" and it's kind of just annoying for the admins to have to enforce a ruling like that just because two people got confused. if we can help it, which we def can in this situation, then we want things like this to be mechanically enforced. if you want it to be a dynamic team like you suggestion, i would consider just making it a lite version of a team antag where people can be assigned a part of the team. which, again, def makes it more of a gamemode (something like this would also help mitigate any metafriending issues; we don't want the same few people doing this every round just because they already know each other). 

> Sorry, I don't mean to be rejecting of criticism I just really want to grasp what the issues are and I don't think I am. Also if every round is too much it could be something that's rolled sort of like a minor antag or even just admin enabled. Though, personally I would advocate for a trial run incorporating it into regular gameplay to see if it really is an issue.

its no big deal, im always fine to explain more about my thought process for stuff like this. if it helps my main point boils down into make it some sort of game mode/antag (the antag dynamic is between the crew and nt rather than the crew and something trying to kill them)

dudeheadface:
> 
> 
> > I don't think I fully understand the issues you're bringing up. I did say in the doc that although the majority of examples given were higher round impact objectives, most actual objectives would be far lower impact like "loyalty testing". I also feel that even the more extreme objectives aren'l likely enough to have the consistent level of impact to warrant being it's own game mode.
> 
> 
> 
> since there are high impact objectives that will need their own rulings and considerations, then we should just treat it like its own game mode. like, think of it this way, DAGD can drastically alter a round, but the fact that it's a lot rarer doesn't mean we don't consider it apart of the traitor gamemode. it doesn't matter that DAGD/high impact objectives are rarer, but we still consider all of them as part of the same gamemode. if we really want to keep something like this as an every round thing, then i think we're going to have to be really strict with what we consider "low impact objectives." like for example i would personally not include the ceo's son objective in the low impact pool, even though it doesn't really Do a lot. 
> 
> 
> 
> 
> 
> > The idea behind many of the objectives is supposed to be that the interests of NT would often make a bad situation worse. Think the one corporate rep in aliens trying to get Ridley and Newt implanted with a xeno after shit already hit the fan. I think the idea of the crew at large being stuck between the threat of antags and the uncaring arm of NT is where this proposal brings the most potential.
> 
> 
> 
> yeah, i agree! i think this is fun and cute. what i'm trying to say is that this shouldn't happen every round. this can be a really common gamerule, that's fine! 
> 
> 
> 
> > As for it not being a full group objective, is the issue that the NTR would technically be the only one that would succeed or fail? Because in practice I thought it would already basically work like a team objective as the only real way for the NTR to complete many of them is by leveraging their position to have IAAs, members of command, or potentially even a few crew members do the actual hands on work. It's sort of a dynamic team objective lead by the NTR as apposed to the exact same roles involved every time which I think is more limiting to potential role play and conflict.
> 
> 
> 
> if it in practice works a specific way, we should codify that to avoid player confusion. if we don't then there is always the chance that someone is going assume "well, im a magi and i only care about space law. fuck nt's interests" and it's kind of just annoying for the admins to have to enforce a ruling like that just because two people got confused. if we can help it, which we def can in this situation, then we want things like this to be mechanically enforced. if you want it to be a dynamic team like you suggestion, i would consider just making it a lite version of a team antag where people can be assigned a part of the team. which, again, def makes it more of a gamemode (something like this would also help mitigate any metafriending issues; we don't want the same few people doing this every round just because they already know each other). 
> 
> 
> 
> > Sorry, I don't mean to be rejecting of criticism I just really want to grasp what the issues are and I don't think I am. Also if every round is too much it could be something that's rolled sort of like a minor antag or even just admin enabled. Though, personally I would advocate for a trial run incorporating it into regular gameplay to see if it really is an issue.
> 
> 
> 
> its no big deal, im always fine to explain more about my thought process for stuff like this. if it helps my main point boils down into make it some sort of game mode/antag (the antag dynamic is between the crew and nt rather than the crew and something trying to kill them)

Thank you for such a detailed reply! I'm definitely getting where you're coming from now. I'll make some edits soon.

PR #59, opened 2026-01-29
[antag suggestion] hitman

stormyatt:
first time doing this so inform of simple mistakes made


stormyatt:
sorry for the mess, did not know how to just replace a change instead of commit

corresp0nd (changes requested):
what does making an entirely new antag add? particularly, why is this its own antagonist rather than building upon the already existing traitor to be better. all of the "kits" you mentioned are just new toys for someone to try and complete a kill objective (which, fwitw, is not the only objective for a reason. i dont know why we would add an entirely new antag just to make kill objectives more common), so like. why not just add them to the uplink store? 

also, anything cosmic cult related is vetoed unless it has to do with the actual game mode. this does not have to do anything with the c.cult game mode, so remove it.

stormyatt:
as described in the second paragraph. The hitman is to fill a roleplay niche for those who want to play a killer role without having to rely on luck or contend with the other objectives a syndicate agent may get. 
as per notable differences hitman 
1. hitman is not aligned to Syndicate or anyone automatically, meaning they work solo. 
2. has a variety of objectives that modify their kill objectives(see, the faction objectives list). meaning they do not just have to kill their targets but do so in accordence to the other objectives.
3. has a random faction per round. meaning the kits they have access to change per round.(and objectives but that already is a per round thing)

corresp0nd:
> as described in the second paragraph. The hitman is to fill a roleplay niche for those who want to play a killer role without having to rely on luck or contend with the other objectives a syndicate agent may get.

i will be more clear then. no, we do not want an antagonist who's goal is to just kill everyone. i don't think antagonists being driven by combat is a bad thing, but in this case, it really just seems like you want an antag with the same overarching goal and theme as the syndicate without the espionage parts. there's nothing here that makes it different from just being like, a type of syndicate agent. even the "no allegiance" thing isn't really unique because you aren't required to work together as a traitor. it's not a team antag

you can do everything in this design doc (particularly the themed kills and special kits) as a traitor with a bit. there's no point in making it an entirely new antag instead of just promoting people to be more creative with the ways they interpret their objectives

stormyatt:
If thats what you think, alright. The objective was just to create a niche for those who wish to play it, similar to thief. but if that seems pointless to you than alright. I assume then that any suggested antagonists that isnt imposibble to be as a traitor is not allowed?

PR #60, opened 2026-02-01
Security Redux

pirakaplant:
Everything you need to read is in the doc itself.

copy_44953B38-F8DF-4489-A4DB-3A0431FADB63

### DIFF
--- a/src/design-proposals/shitsec.md
+++ b/src/design-proposals/shitsec.md
@@ -2,11 +2,11 @@
 
 | Designers | Implemented | GitHub Links |
 |---|---|---|
-| pirakaplant | :information_source: Open PR | funky-station/funky-station#2828, funky-station/funky-station#2829 |
+| pirakaplant | :information_source: Open PR | *(Funky)* funky-station/funky-station#2828, funky-station/funky-station#2829<br>*(Forky)* funky-station/forky-station#27 |
 
 ## Overview
 
-This is a proposal to remove all nonlethal options from security's arsenal. They will have (nerfed) truncheons as a "less-lethal" weapon instead of stun batons.
+This is a proposal to remove all nonlethal options from security's arsenal. They will have nightsticks as a "less-lethal" weapon instead of stun batons.
 
 ## Background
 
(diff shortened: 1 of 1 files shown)


taydeo (approved):

pirakaplant:
> And maybe it's a case of personal preference but I vibe more with the idea of shock-based weaponry for sec, kind of invokes the feeling of the Combine from Half-Life (Pick up that can), brutal but not too brutal as that is going to leave a mess that later needs to be cleaned up

I think they should make a mess, actually. I don't want to pull any punches with Funky's themes.

> Assuming we went with the idea of giving disablers shock damage I would consider calling them "Less lethal" to only be borderline correct, i would consider shock-disablers to be proper weapon, just with its own niche to fill, especially now with disablers only dooing slow down, that means a typical arrest would likely involve many disabler shots to a target before the secoff closes in with a Stun button/Truncheon, in either case that's going to leave the arrested target pretty heavily beaten up.

I think that if you run from Security, you should not just expect to be injured, you should expect to die. Again, I don't want to pull punches, and it makes arrests a lot more tense than just getting an extra three minutes for running away.

ferynn:
I think this idea is good. Though I'd recommend not doing half truncheon damage, as that's only 10 blunt. That's less than a harmbaton currently.

taydeo (approved):

corresp0nd (approved):
theres been plenty of discussion about this doc already and yes, i am on board. i dont have any qualms with the doc itself or the idea. one thing i would like to see more thought put into is how stamina damage works. "the gun that harmlessly knocks you over" feels like shit because the way stamina damage works is shit. the stunmeta feels inauthentic because the stamina system and the health system are two entirely different things--able to be healed and dealt entirely separate of each other. theyre two bars on your video game screen. reflavoring sec's arsenal isn't a bad thing, which personally seems to be the primary goal here, but i would like to see a more indepth look into these systems entirely

radape:
Hi! Was redirected here after some malding on the game discussion chat so imma try and put my thoughts into a more coherent form than just crying because sec killed me or something.

I really do like this change, and the direction funky is taking things into as a whole, it's good that y'all finally removed this weird dissonance between sec being the police force of an oppressive megacorp and gameplay wise them acting very non-lethally. But my main worry at the moment, both from my own experience playing these changes and the general reaction from people, is that it feels like it's _too_ easy get away with being shitsec?

Part of this is just the PR being like 3 days old and people needing to adjust, using flashes and bolas if they wanna take someone "non-lethally" over just batonning them into crit as their only option, but part of it is that there's already two systems in place that are supposed to help keep sec in some sort of check, and it's too easy (and honestly, just better overall for the round) to ignore them: lawyers and CC officials (IAA, NTR, Magistrate).

NTRs and IAAs are already gonna lean more on sec's side, but even if they don't turn a blind eye, enforcing SOP is hard when you can't dish out punishments for it in a reasonable timeframe. An officer beats someone to death? IAA has to learn about this, contact the HoS which more often than not may ignore their reports while dealing with a threat, escalate chain of command, and if there is an NTR or Magistrate, hope they issue a demotion (as they can't do it themselves), which will take an extra amount of time as it requires coordinating with the HoS, captain, and the offending officer. It's good RP, but it's a big RP time sink to deal with one or two infractions, which a lot of players don't seem to want to engage in (or when they do, it becomes a slog of waiting for other players to be available and willing to buckle down for some talking RP).

Lawyers as a whole just really depend on the Magistrate being there to officiate a trial, and beyond that they can't really enforce anything besides being enough of a nuisance that whoever they're dealing with relents. It's kind of how the job works, it's pure RP, I genuinely don't have a clue how to do something about it.

End of the day this is just my own take on things, but I do like the idea that sec going too far too easily should have consequences. They are impeding other departments from working, after all, and the crew can't really work under threat of nightstick forever, there's incentives for NT to want to keep things calm.
I feel like Unions will end up taking a big chunk of this job by allowing the crew themselves to dish out consequences in the form of strikes, but I wonder if it'd be good to consider some ways for CC officials to be able to dish out more tangible punishments on a department and for command to handle demotions more easily, maybe even be incentivized to "get rid of the problem" temporarily much like we see in real life cases of police overreach. Send them on a forced vacation, essentially, to keep the façade up and appease the station.
In a more general sense sec will still be winning, but in the context of a round, a player will still be demoted. This would also have the bonus of cementing sec as another part of the NT machine: they're the ones with guns, but they're beholden to NT end of the day.

I am spitballing though, I'm sure someone more in tune with funky's design could make something better from this, but all in all I'd like to see more iteration on this idea of bringing Sec closer to real life police forces. I know the frustration is the goal, but I hope more ways to work with the frustration can come up from this.

PR #61, opened 2026-02-02
nanotrasen retheme/rebrand 

corresp0nd:
tldr; reconsider why the station is even there and wtf nanotrasen is doing. i hate the current approach of a vague "research" so this rethemes nt to be a defense contractor corpo (and also considers all the ways the game needs to change in order to actually communicate this) 

pleaseeee review this but it is still a wip. this is my first draft sooooooooo i might rewrite it (and i need to expand on the shit i didnt)

### DIFF
--- a/src/design-proposals/nanotrasen-rebrand.md
+++ b/src/design-proposals/nanotrasen-rebrand.md
@@ -1,0 +1,68 @@
+# NanoTrasen: Defense Contractor and Industrial Corporation
+
+| Designers  | Implemented | GitHub Links |
+| ---------- | ----------- | ------------ |
+| correspond | :x: No      | TBD          |
+## Overview
+
+So, what the hell is NanoTrasen doing out there anyways? 
+
+Generally, the acceptable answer is a vague *research.* Depending on who you ask it's research about plasma. The station isn't there to do much of anything except kind of exist, meaning that players and developers can just do whatever they want. There's no reason for characters to be out there, except to do their job (whatever that means).
+
+For the most part, this is just fine. It's not bad, many corporations do just push papers around to fulfill some imaginary quota. It works out that a lot of the bad things that happen to the station can be explained as a vague type of insurance fraud. NanoTrasen doesn't give a fuck about the station, so being able to get rid of it "accidentally" and an insurance payout is great for them. 
+
+However, it leaves a lot to be desired. As a player, you don't know what that "research" is even supposed to be about because none of it matters. The stuff science researches doesn't even have anything to do with plasma. You have to ask around, outside of the game, to have an idea about what it's all about. The insurance fraud tidbit is like this, too. 
+
+Even if we did try to integrate the reasoning into the game more, it would still kind of suck. What does research mean? What's the point of the research? We can keep explaining this all as a form of corporate fraud, but that's kind of just lame. Nothing *has* to change if this is all just a vague form of fraud. 
+
+Instead, we should consider reframing what NanoTrasen is doing there in the first place. NanoTrasen is an intergalactic defense contractor and industrial corporation, the station being one of the satellite research and development plants. They primarily focus on plasma-based weaponry (this should personally lean more sci-fi, but it doesn't entirely matter what kind of weapons they manufacture). This all being a really wordy way to say that NanoTrasen is Lockheed Martin or Raytheon. They're glorified arms dealers who also love to do fraud. 
+
+The primary goal of redesigning NanoTrasen like this is to reframe the entire game. Characters have a reason why they decided to work *here* specifically. It is non-negotiable that players have to interact with aspects of the world that make them uncomfortable. By forcing people to consider why their character would choose to work in the military industrial complex, we're setting the tone for the game entirely. This should, hopefully, help alleviate some of the issues we have with players not interacting with certain tones we want to encourage. 
+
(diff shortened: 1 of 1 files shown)


ferynn:
I do think tying sec more to a Nanotrasen military theme would be good. It’s the sovereign entity out there, and it would be better to drill home that space law is Nanotrasen’s law they’re applying

SophiaImpetus:
My thoughts on how this world would all work.

In the setting, and especially in the story that it seems we want to tell, if Nanotrasen is a huge techno-feudal colonial power, then what does, or doesn't, have value is going to be expressed differently.

What has value to the shareholders of a corporation large enough to own many solar systems and regions of space isn't goods or profits. They have a large enough power base that if anyone who is a beneficiary of the regime wants it, they could have a 100 story palace built on a private moon in the matter of a few months and this wouldn't be a blip in the resources they have available to them. Most of the galaxy that isn't a black void is mineral wealth, materials you can manufacture anything out of. In a post FTL galaxy, the value of raw materials goes down significantly. 

What has value to the shareholders is property, and power. A "Research" station needn't discover anything, nor be dedicated to development at all. As corporations grow even more powerful, and desire more and more types of property, information becomes a type of property too. The first corporation to extract some kind of research data owns that data forever. Research here is just farming information, using assets that have little value (a space station, workers), to produce something with greater value to the shareholders, more property under Nanotrasen's control. Every readout from an anomaly or artifact represents something Nanotrasen now controls that another corporation can't use or exploit without paying them the price, if they let them use it at all. More leverage. More power. Every last one and zero of data harvested from the readouts are patented and made into Nanotrasen property. 

Nanotrasen expands into space not aligned with TSF using its military power because owning more solar systems increases the value of Nanotrasen to shareholders. Nanotrasen embraces war and selling ships and components to TSF because in a world where resources are so low in value, nothing burns resources and worker power better and increases their value than war. As a result then, workers are the real asset of Nanotrasen. You can have all of the resources in the galaxy, and yet what you can do with them is limited by if you have the people to do it. As Nanotrasen grew in size securing workers this way would have been an obvious next step.

The economy Nanotrasen's citizens work in is entirely fake, it's a play economy built because they know that creating a system where people can't live anywhere, eat, drink, without spending scrip they have to earn from Nanotrasen is a system where they have all of the labor they want. TSF probably wouldn't align with them if they outright enslaved their workers (with exceptions for borgs of course because robotic life is an "obvious" safety risk worth "mitigating" with laws), so running a fake economy like this is the next best thing. In fact its even better, because if people were enslaved, they'd be more likely to revolt or question things. Whereas when people are competing with each other, and feel in control of how well they do, they're more likely to be compliant. Unless one of Nanotrasen's worlds was in the center of a real armed conflict where supply lines get cut off, space operations are prevented by gunships and real scarcity sets in, anyone who lives on a Nanotrasen world could live in relative comfort and abundance (from our current time's perspective). But none of these people's lives have any meaning to the shareholders and ruling class of Nanotrasen except in terms of what value they produce to the company. If they're all starving, struggling to survive, it doesn't matter. Resources were saved and they are motivated to work for less. They're not being "entitled" to the profits of a company they don't own, to the profits of their own labor that they don't own. That's a good thing. Most characters won't have a reason why they decided to work in Nanotrasen rather than anywhere else - because everything in their lives is already going to be Nanotrasen, either from the big corporation itself or one of billions of subsidiaries right down to the little donut shop a entrepreneur on the station is running to take some of your hard earned scrip. Nanotrasen isn't really hiring people, it is putting them to work so they can pay them in their fake currency to gain things that actually have value for the corporation.


That's basically why I don't think the station needs to be "developing" anything and also why I don't think it needs to be manufacturing anything. Its going to be more expensive to manufacture in space than on the land, unless the purpose of the station is to manufacture warships, in which case yeah it would take place in space (not saying that wouldn't be cool as fuck either but I dunno if we want to do that). Researching in space makes sense though - if something goes disastrously wrong, wiping the problem away is so much easier, and collateral damage is pretty much an non issue as well. 

I'm all for killing the lathes and research unlocks though. Personally rather than manufacturing things on site being the way of getting advanced products, I think it would be much more flavorful to force people to engage directly in the fake economy nanotrasen is running, buying advanced tools and anything else they need with scrip or spesos. The [Cargo Rework Document](https://github.com/funky-station/docs/pull/65) provides much more meaningful flavor around the kinds of supply lines and traders than just teleporting things in, and I would like the supply department to actually be the backbone of station supplies.
In my opinion the thing the station exists for (harvesting research data for Nanotrasen) should be the thing that the station is getting paid for as a subsidiary of Nanotrasen by its parent company. Science wouldn't produce research points or unlock technology but produces spesos as a bounty for the data they are transmitting to Nanotrasen. The supply team only needs to bother themselves with spending it, unless an opportunity arises and they want to make some more money, but they wouldn't be where the station budget comes from at all.

A station that has manufacturing focused goals would be cool too, to be clear, but I just don't see a need to keep science as a department at all in a fully military manufacturing station. If they're out in space shipbreaking, mining, using those materials in building warships and then fully kitting them with the requisite supplies and guns ready to be sold to TSF or another part of Nanotrasen, why would things be ordered to be manufactured that don't technologically exist yet. Why would the technology needed to make plasma rifles or something not exist yet until you poke an anomaly a few times. You might as well do away with science and properly focus on a new purpose for the station.

PR #64, opened 2026-02-07
Requirelidae (Requir)

MoxieIsMoxie:
It's polecat time (this time with less 99% main)

### DIFF
--- a/src/design-proposals/Requirelidae.md
+++ b/src/design-proposals/Requirelidae.md
@@ -1,0 +1,75 @@
+# Requirelidae 
+
+| Designers | Implemented | GitHub Links |
+| ----- | :---: | ----- |
+| Moxie\_is\_Moxie, literallyablueberryguy | ❌ No | TBD |
+
+## Overview
+
+This doc adds the Reqirelidae, a polecat-like species that has a quick circadian rhythm.
+
+## Background
+
+With many species being consuladated, the opening for new species has begun with many people wanting more mechanical depth and interesting species. 
+
+## Features to be added
+
+### Requirelidae (Requir)
+
+ruh·kwai·luh·dai
+
+#### Lore:
+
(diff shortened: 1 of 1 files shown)


corresp0nd (changes requested):
the lore is... whatever. im always going to be pretty iffy on a species becoming humanoid through genetic bioweapons/alterations/whatever. its a kind of bad way to explain why a species is being treated the same way as other humanoid species. it frames the world in a very ethnocentric way; something something the only way its possible for other species to become "as good as humans" is by humans doing something, rather than it just Being Possible for other humanoid species to exist. i would strongly suggest you cut this aspect of their lore, especially because it doesn't really matter. like, they can just be susceptible to disease and genetic issues because they're from an entirely different galaxy where 90% of diseases don't exist. 

the active/dormant cycle is kind of cute but im not really sold on it. i dunno. i dont think its a bad idea and imo it does warrant its own species to be made, but its just like. Eh. it seems Ok and Kind Of Interesting. i would maybe like to see how this could impact their gameplay or something to be more convinced on it. 

the rest is fine, but probably cut the night vision and crawl speed. i dont think they need it.

MoxieIsMoxie:
> the lore is... whatever. im always going to be pretty iffy on a species becoming humanoid through genetic bioweapons/alterations/whatever. its a kind of bad way to explain why a species is being treated the same way as other humanoid species. it frames the world in a very ethnocentric way; something something the only way its possible for other species to become "as good as humans" is by humans doing something, rather than it just Being Possible for other humanoid species to exist. i would strongly suggest you cut this aspect of their lore, especially because it doesn't really matter. Like, they can just be susceptible to disease and genetic issues because they're from an entirely different galaxy where 90% of diseases don't exist.

They were always humanoid; the bioweapons just semi-warped them to what they are now. It was a bioweapon that they inflicted on themself which eventually forced them into space before they were ready for full space travel (like if we tried to fully colonize space before we even landed on the moon). The reason they developed the cycles is that they left ships that needed 24/7 maintenance, and there was little time to sleep. All of this happened long before they made contact with anyother specie. If you feel as if this is still too much, I can reformat it into another ecological disaster. 

I will expand on cycles and definitely cut night vision.

PR #65, opened 2026-02-07
Cargo Rework

Joaco545:
Adds a design document that talks about reworking how cargo currently works by adding traders to the game


Ratsmageddon:
I do want to note (for better or worse) this'll make sabotaging cargo by like, nukies, a lot harder. Even if you total the cargo shuttle, you'll need to total the salvage shuttle, which, unlike the ATS, will probably make you caught if you're trying to stealth sabotage.

It's mostly just a balance change on that part, and I don't know how big of a change such would cause.

GansuLalan:
I like it, and it matches up a lot with my own thoughts on station side cargo I remember trying my own hand at shuttles a while ago, also based on the goon design, but found I hated UI so much I shelved it lol. My main consideration would be around annoyance, the doc suggests that cargo can only sell to traders to regain annoyance but from my reading a too high annoyance would result in no more traders and therefore no more money income with no way to recover? Crippling a department is fine but it should always be recoverable in some way so there would ideally have to be some always available method to lower annoyance even if it is slow which it probably should be.

My other consideration, though I don't know if this would be something to solve in this design doc, is around what people tend to order at the moment which seems is not much. In my experience, assuming the average round, the only people who order stuff from cargo is janitors, botany, and chemistry and then generally plastic and textiles with other materials handled by salvage, maybe I just have quiet rounds though. Outside of that emergencies are the big rushes and that would probably become awkward if the relevant trader is missing and you've got to wait however long for the delivery.

But ye, overall I'm very happy with it, I fear the amount of work that will be required but it is all worth it for the dream of killing the bounty console. I've been told to produce a design doc for cargo and while I've taken that to mean more about the theming and what to think about when designing for the department but this covers most of what I wanted gameplay wise for the cargo tech people.

Joaco545:
First of all, thank you all for your comments! Second, sorry for the wall of text as I reply to all 😉 


> [...] it mostly felt like "grab x items to sell" be it bounties or something expensive like charged smes, it always felt like just grabbing some random items for no real reason other than to sell.

The SMES farm... good point. Added an idea for an item market. Sprinkles a bit more randomness at the start of the shift to move the meta of what items to sell, and also linked to this system, prices get lowered if something gets sold repeatedly.
A smart Cargo Tech/QM would learn how to play the market, which adds more depth to the cargo department too.


> I do want to note (for better or worse) this'll make sabotaging cargo by like, nukies, a lot harder. Even if you total the cargo shuttle, you'll need to total the salvage shuttle, which, unlike the ATS, will probably make you caught if you're trying to stealth sabotage.

Traders add an inherent delay when ordering things through NT's systems, as the delivery ship would take time to get to the station. As such, I think that actually makes it a lot more favorable for nukies that want to suddenly (blitz) or stealthily strike; Maybe too favorable... Depends on the timing used.
Unless a trader who sells weapons happens to pass by, ordering on the cargo console would have a delay that would probably take longer than defending from the nukies.

Maybe an express delivery option can be added for a transport fee... Still no instant weapons on the ATS, but faster from a normal delivery

Thanks for bringing this point up, @Ratsmageddon. I had not considered the need for rapid access to more weapons.


> My main consideration would be around annoyance, the doc suggests that cargo can only sell to traders to regain annoyance but from my reading a too high annoyance would result in no more traders and therefore no more money income with no way to recover? Crippling a department is fine but it should always be recoverable in some way so there would ideally have to be some always available method to lower annoyance even if it is slow which it probably should be.

The "they will stop visiting the station if they don't have a purpose" part of Annoyance was meant as a "Trader ships will still go if they have to make a delivery to the station". Combined with "and decreases when favorable deals are struck with the traders" the idea is that when traders refuse to come to the station unless they have something to do, cargo still has the option to buy things over market value and complete bounties under what they are asking to be able to reduce annoyance.
Though I would not expect players to get to this point, as getting there would mean annoying a lot of traders a lot consistently throughout a shift.

I will see about rewording this part a bit to make it a bit clearer that there is always a path to recovery, albeit a painful one, as the station will lose profits (and NT doesn't like that. Admeme possibility for Epsilon?)


> My other consideration, though I don't know if this would be something to solve in this design doc, is around what people tend to order at the moment which seems is not much. In my experience, assuming the average round, the only people who order stuff from cargo is janitors, botany, and chemistry and then generally plastic and textiles with other materials handled by salvage, maybe I just have quiet rounds though.

I am kinda counting on Scrip and Scrip-based shops to add more volume for item deliveries. Else it would fall on mail deliveries to have cargo moving semi-consistently


> I fear the amount of work that will be required

You and me both... This will probably need a semi-large workgroup if we want it completed before 2030 😅

corresp0nd (approved):
this is REALLY cute and a great way to overhaul a department that much needs the love. im completely on board with this personally. i particularly really like the minigame for the ftl comms computer, it seems super fun and cute. i think my only concern is how we might handle the haggling minigame, i think it needs a slightly more in depth plan before we go for it. but also imo that is a later issue and we can just take it as it comes lol

SophiaImpetus:
I just had an interesting thought. 

Since you need to wait for things to be delivered after you order them, an important new job for the Quartermaster could be deciding when to hit the button that calls a ship to deliver items ordered directly from Nanotrasen via scrip or spesos, with the delivery having a substantial fee in spesos for each time the ship is called in to visit. Until then approved orders will just be prepared at whatever planetary warehouse is stocked to ship them out to you.

Joaco545:
The way I was thinking about this is that when you buy things, they get queued, and when a trader ship intends to go to the station, it would pick up the orders and MAIL* and then jump in, making a ship go to the station about every 15 +- 5 minutes with whatever was ordered.
I think NT would be cheap enough to pay some minimal fee themselves, or have a stipulation in the trader's contract that allows them access to NT space, which would make them deliver things as they are already going to the station

Having said that, having a way to manually force a Nanotrasen Delivery ship to immediately deliver for a fee (express delivery) would be something NT would have as an option. Gameplay-wise, you don't have to risk traders taking too long to come in during an emergency if you make sure to have an emergency fund (QM SOP?)

*To emulate how the current mail system works, ships would need to come periodically to deliver mail

empty0set:
@Joaco545   I've read the doc and think its a pretty cool direction to take cargo in with a lot of potential.  However, there are two things to consider here:
1) for large projects you should have a minimum viable product (MVP) in mind.  Like what is the most basic, bare-bones aspect of this that can still be a functional gameplay loop.  What comes to my mind immediately is replacing the ATS with supply ships that FTL in with purchased goods + mail (sans the traders themselves) as a starting point.  Doing something like this will allow you to prioritize tasks for development & allows for flexibility for any unforeseen issues that may arise.  Without this you run the risk of building castles in the sky so to speak--creating more and more features which, while nice to have, are really contingencies upon contingencies when whats really needed is a solid foundation to iterate on.
2) Once a starting point has been identified, every feature needs to be broken down into concrete tasks for them to be implemented.  I tend to think of things sequentially, like an order of events as they would occur in a real round.  Returning to the MVP example, I would break this down like:  Order placed on Cargo Console -> NT delivery ship is spawned -> NT delivery ship leaves after some time.  Now flesh it out in granular detail, try to enumerate edge cases, or different paths.  For instance, how is the ship spawned?  Near the station?  It could probably re-use the code for spawning ghost shuttles.  Near Cargo specifically?  You are probably going to have to write that yourself.  How are the bought items spawned on the shuttle?  Code would have be added to spawn these items on the shuttle when loading the grid.  What if some items cannot be spawned (for instance if there isn't enough space)?  A queue of bought would need to be maintained.  Etc.  Etc.  Again this is just an example, but I hope this illustrates the process of concertizing these ideas into something that is much more approachable in terms of implementation.

Joaco545:
Thank you for you tips @empty0set! I do indeed have a vision for an MVP. I have a [second doc](https://docs.google.com/document/d/1cvGWMJNLofQUpYoUlqbQM4iz7LiguteZKmAMjIsCrA8) I kept for myself (sharing it now if you want to peek) that tries to specify the needed systems and mechanics that need to be implemented a bit closer to the ground. This is a live thing that grows when I think about how to implement this project, and it's also a space I want to share with anyone who wishes to help making Traders a reality

Now for specifics, the MVP will have what I call the "Trader Management System", which handles the spawning of ships (gonna re-use ghost ship spawning) and station orders (keeps track of what was ordered), and the "Delivery Ship Mechanic", which handles the delivery of requested items and mail.

As for the spawning of cargo you mentioned on the second part as an example, the Doc states I plan to reuse and modify the ATS's cargo spawner platform, tied to a queue that keeps track of what needs to be spawned next. While anything here is subject to change, I decided on this method for easy implementation potential, infinite scalibity potential (spawning things on the moment kills the "not enough space" errors), while also being able to keep set limits if wanted (eg cargo capacity can be a YAML edit for the ship type/model).

Also, I will show my software development student card here :P
While I do have much to learn, project milestone setting is something I luckily understood early on... *Looks at pile of unfinished prototype projects*... Ok, maybe it needed a bit more trial and error to get right, but I have it now!

PR #66, opened 2026-02-07
[Antagonist] The Voidwalker

TrixxedHeart:
Design proposal for The Voidwalker, a mid-round ghostrole antagonist based on TG13 focused on abducting crew.

image


corresp0nd (approved):
seems cute and fun, i particularly like the note of the radio jammers being the way for crew to "fight" back without just like, gunning the thing down. i think it adds a lot of tension that works out really well for a midround. being permanently affected by the voidwalker's transformation is also a really fun thing to see, i really like the direction this is going!

PR #51, opened 2026-01-21
Judicial Process Changes

TheBoss913:
There was a convo about expanding on the Judicial Process, and this is something maybe.

PR #67, opened 2026-02-08
Court and Trials Reworked

ferynn:
Rework of the criminal sentencing system to spotlight trials, and lawyer/IAA/Magi interactions

See proposal doc.

Give comments pls

### DIFF
--- a/src/design-proposals/courtandtrials.md
+++ b/src/design-proposals/courtandtrials.md
@@ -1,19 +1,15 @@
-# Short, Properly Capitalized Title
+# Court and Trials
 
 Reworking the Legal System
 
 | Designers | Implemented | GitHub Links |
 |---|---|---|
-| Ferynn + Dev Chat | :white_check_mark: Yes or :warning: Partially or :information_source: Open PR or :x: No | PR Links or TBD |
+| Ferynn + Dev Chat | :x: No | PR Links or TBD |
 
-`Designers` should be the names that you use on GitHub and/or Discord. This is optional but strongly recommended, since:
 
-- This acknowledges credit where it is due
-- People who are confused about the written intent can use this information to contact the authors
+See this discord conversation for relevant parties
 
-`Implemented` is the status of the feature.
-
-Github links can include multiple PRs, if relevant.
+https://discord.com/channels/1276640157511979008/1276653734767755294/1460819328650706956
 
 ## Overview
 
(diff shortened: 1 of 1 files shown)


src/design-proposals/courtandtrials.md
```diff
@@ -0,0 +1,87 @@
...
+
+While charges are recommended by the prosecution, and they build their case to support them, all final decisions on charges are given by the presiding judge. If guilty of any crimes, the accused is sent to the brig for their sentence. If not, they are let go.
+
+A prisoner can waive their right to a trial and enter a guilty plea, to expedite the court process. While this does not always lead to reduced sentencing, the presiding judge should take cooperation into account. This allows for the creation of plea deals between the defense and judge, as well.
+
+### Required Dev Work
+
+To support this system, at minimum, a PDA app would need to be created, allowing for relevant parties to be pinged when a trial is imminent. A sample default set of PDAs including this would be the Warden, CC VIPs, Lawyers, and the HOS. Usage of a software cartridge would allow for this system to be included for a chosen Baliff. 
+
+This could be potentially expanded with further work into a more advanced system, so information could be transferred on the impending trial to relevant parties, and allowing for information to be sent back to the warden on sentences.
```
Joaco545: And lastly, I think players can handle this without needing a special app on the PDA.
Nanochat and an intercom in the courthouse should be more than enough

Joaco545: Yeah, exactly. A nanochat group for the court would fit this niche perfectly. As long as the bailiff or judge announces on the chat that the court will start soon, everyone will be notified

corresp0nd (changes requested):
my concerns remain largely the same from #51, i really just don't see trials working unless we decide to go in a very drastic direction with them. we can add SOP like this to solidify what is supposed to happen, but without any mechanical backing, i think they'll mainly just be ignored. i don't think this is necessarily the fault of the players here for refusing to interact with a system like this, but more so that it kind of sucks when 90% of cases are just Not going to go the way that we want them to. players will try it out at first, realize it sucks, and then dread playing this specific part of the game so much that they avoid it. and trials are something where Everyone involved Needs to interact with it to work. 

for example. we want trials to be like, 10 minutes max, sure. we can both agree on that. but if we add a stipulation like "the time to get to the courthouse can be extended if there are extenuating circumstances" then 90% of the rounds are going to have some sort of extenuating circumstance. that's the pitch of the game, we're playing a disaster sim. trials objectively cannot be something super structured (where the chaos comes from inside the courthouse rather than outside of it), but we also need that really strict structure for them to remain a footnote of the game and not the entire gameplay. 

hopefully this better explains where im coming from when i say i dont like trials are going to work. i think, theoretically, we can do Something with the idea, but we need to be a lot more drastic with what we even consider "court gameplay" in the first place. i also do not believe this is something that can be achieved through just SOP and paperwork structure, even though that is what people like about these roles. there needs to be some sort of strong, mechanical system backing these types of systems, even if they don't necessarily seem like they should have a system like that.

ferynn:
I think most trials could get done within a minute or two honestly. Most crimes are pretty open and shut. I do share the concern on the time of them potentially dragging on.

I don’t agree though about the first bit. About things being ignored. Trials currently are an optional system for long drawn out appeals basically, currently. 

Something like this is different in that it entirely shifts the job of who assigns sentences. People would have to interact with it in at least minimal capacity, as the Warden simply wouldn’t be allowed to sentence anymore.

Even if it was just 30 second rubber stamp trials, some effort would have to be made to shift towards courts and procedural process for crimes. They wouldn’t be able to just ignore it and do something entirely different.

pirakaplant:
Looks good to me! No notes.

PR #68, opened 2026-02-11
ID & Contracts

MoxieIsMoxie:
This doc is to expand on some of the ideas of Unions and Scrip docs and overhaul some current systems to pull them into a more unified vision before attempting implementation.

 Let the class disparity begin.


corresp0nd (approved):
this is really cute and seems like a lot of fun, plus it pretty much perfectly fits into the type of world we're looking to try and communicate. i honestly don't have a lot to say about it, though when you start implementing some things you might run into some nitpicks from me. i would work on the security ticket and scanner stuff later, though. i think there's a couple other folks who have their own takes on the direction security should go and it's probably good to talk with them about stuff like this

PR #69, opened 2026-02-12
CT Scanner Machine

vivry:
This is a proposal for the addition of a CT scanner to med. It will mostly replace the functionality of the health analyzer. I will start implementing this soon, so let me know what you think about it!

### DIFF
--- a/src/design-proposals/scanner.md
+++ b/src/design-proposals/scanner.md
@@ -1,0 +1,62 @@
+# Medical CT Scanner
+
+| Designers | Implemented | GitHub Links |
+|---|---|---|
+| vivry | :x: No | TBD |
+
+## Overview
+
+This proposal aims to introduce more depth, complexity and RP potential to the medical department by adding a new machine that loosely replicates a CT scanner from real life. This machine would replace the functionality of the health analyzer, providing an exhaustive outline of a patient's status.
+
+## Background
+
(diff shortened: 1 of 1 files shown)


corresp0nd (changes requested):
i think this is cute and a good take on health analyzers, but before you require the usage of it you GOTTA improve triage gameplay. it's fine that we want to slow down medical gameplay a lot like this, but in high stress situations this is just going to piss people off and cause a lot lot more medical mispractice. there needs to be a way for medical to say "Ok this person is about to die they are going to get epi and a stay in the super fast ward" versus knowing that like, someone just has an upset tummy and 10 poison damage. and people need to be able to communicate that difference near instantly because every second counts in this type of situation.

vivry:
Yeah, on closer inspection I agree with you. I was already a bit worried about the pace of treatment itself, which is why I'm also making some amends to try and soften the blow. When I was writing the doc, I assumed that triage could be done just via the inspect menu, but it may not be descriptive enough to provide effective results. I was considering some ways of making it faster. Which approach do you think would be better for improving this? Let me know what you think and I'll edit the doc for re-review.

1. Improving the inspect (shift-click -> health) menu, especially by providing visible signs to exceeding high thresholds for all damage types.
2. Instead of making health analyzers restricted to CMO, making it so that they only provide the total amount of damage on someone, without any further specifics. This lets med triage people by just clicking on everyone and sorting them by the biggest number, but without giving away the actual course of action for treatment. Knowing what to treat would still require asking the patient in simpler cases or using the CT scanner in more complex cases.
3. Same as 2, but instead of the sum of all damages, show only what the highest damage value is, without mentioning what its type is.

corresp0nd:
i think option 2 would be the easiest to help integrate people into this workflow, but tbh it doesn't Need to be the med scanner at all. i think you could use the medhud for the same purpose or make something entirely new, if you want. either way i don't think any of the options are Bad directions to go in, so it's really up to you

vivry:
Alright! Using the med hud is a good idea, but what would I need to change for it to make triage better for medical staff? I tested it and it already shows someone's damage at a glance, as well as whether they're critical or dead. Now that I think about it, it already performs this exact function of quickly letting its user see the total damage on someone. It just doesn't show the actual numbers, but it effectively does the same thing by visualizing how close someone is to being crit or dying.

corresp0nd:
yeah.... probably.... idk jsut go for it

corresp0nd (approved):

PR #73, opened 2026-02-14
Quylon Species

SophiaImpetus:
Quylons are a lithoid species conceptualized to be a severely challenging pick for players, but stand out in uniqueness and contribute to a more authentic sci-fi and occult atmosphere on the  station.

Currently a draft PR as I feel there are a lot of potential issues here that need to be thought over and resolved before considering it a finalized proposal. Will a scavenger hunt be too time consuming or demand too much attention in each round for instance? @MoxieIsMoxie suggested changing this to a more simple goal of performing the ritual at a randomly selected location on the station, which may still pose some obstacles without being as extreme. Would this be preferable?

Comments and feedback greatly appreciated here.


Terkala:
What niche does this species fill? It seems to have some pretty massive upsides for antagonists (ie: nukies who are functionally immune to lasers), with a downside that is either completely irrelevent, or entirely crippling, and the two cases may be out of their control. Balancing something with a massive advantage and a massive disadvantage is always going to be tricky.

This could mean that a command member of this species would just be doomed to die at the 50 minute mark if they get an unlucky roll of what items they require.

I'd say to ditch the attenuation and heat resistance entirely. Lacking hands (and insuls) and boots is already a massive drawback, that justifies the unslippable upside.

SophiaImpetus:
> What niche does this species fill? It seems to have some pretty massive upsides for antagonists (ie: nukies who are functionally immune to lasers), with a downside that is either completely irrelevent, or entirely crippling, and the two cases may be out of their control. Balancing something with a massive advantage and a massive disadvantage is always going to be tricky.
> 
> This could mean that a command member of this species would just be doomed to die at the 50 minute mark if they get an unlucky roll of what items they require.
> 
> I'd say to ditch the attenuation and heat resistance entirely. Lacking hands (and insuls) and boots is already a massive drawback, that justifies the unslippable upside.

The gameplay niche is supposed to be that species which add additional challenges to each round are fun, and there is a lack of them on funky. The roleplay niche is having a much more alien feeling species than is currently available. 

It is fair to say that the current design outline has gone too far in both respects.
Stuff like the heat resistance is just flavor. In hindsight it doesn't even make sense, with blood that can boil and fragile crystals that might warp under heat. And yeah given the game balance concerns I will change it. Fire immunity might be more reasonable?

Attenuation I feel is integral to the flavor of them, removing it entirely would remove a significant part of the gameplay which makes the roleplay tangible in game terms. I do find the danger of it fun, I think having a magical recharge system is cool, but it doesn't need need to be expressed in such harsh terms as psuedo round removal. Alternatively if it is expressed like that, maybe the requirements to de-attenuate is the issue. Or both sides of the equation need to be changed.

PR #74, opened 2026-02-19
HUD Removal (& the Panopticon Helmet)

pirakaplant:
Everything you need to read is in the doc itself.

HAz9_KiaAAEc3me

### DIFF
--- a/src/design-proposals/hud-removal-panopticon-helmet.md
+++ b/src/design-proposals/hud-removal-panopticon-helmet.md
@@ -1,0 +1,83 @@
+# HUD Removal (& the Panopticon Helmet)
+
+| Designers | Implemented | GitHub Links |
+|---|---|---|
+| pirakaplant | :x: No | TBD |
+
+## Overview
+
+This is a proposal to remove all HUD eyewear from the game in favour of existing, non-HUD eyewear as appropriate. Warden will get a helmet that helps them surveil the brig instead, for gameplay and aesthetic reasons.
+
+## Background
+
+Lately, there has been a lot of discussion concerning the direction we want to go in, both aesthetically and mechanically. Security glasses have been brought up in tech-level discussions in #lore as both something that does not fit our aesthetic and an example of QOL slop that negatively impacts roleplay.
+
+When writing this document, I took a look at the other HUDs, and realised that none of them actually justify their own existence either (see Game Design Rationale).
+
+## Features to be added (well, mostly removed)
+
+- The security glasses and its variants will be removed from the game.
+- The medical HUD and its variants will be removed from the game.
+- The diagnostic HUD and its variants will be removed from the game
+- The administrative glasses and its variants will be removed from the game.
+- The beer goggles will be removed from the game.
+
+Instead:
+
+- Jobs that had security or administative glasses instead get sun glasses (except HOP since their version didn't have flash protection to begin with).
+- The Warden will recieve a panopticon helmet (see below).
+- The "table sliding" ability of the beer goggles would instead be relegated to "bartender gloves" or something similar.
+- Sunjar glasses will be an option to aesthetically replace the security jamjar glasses (and also provide more options to the captain and other jobs which had admin glasses).
+
+### Panopticon Helm
+
+A bulky, white helmet with an opaque, adjustable visor that covers the eyes. It hums like a distant, untuned TV. A lopsided radio antenna sticks out the top. An icon of an eye is stencilled onto the front.
+
+The panopticon helm can be found in the Warden's locker. It can be worn in the helmet slot. It slows the wearer down slightly and provide mild protection from brute damage.
+
+As long as the eyewear slot is free, the wearer can lower a viewport over their eyes to tap into the Security department's cameras (either through the context menu or the action bar, like a hardsuit's helmet). This provides flash protection and security HUD vision (minus mindshield status). It also removes line-of-sight, but only renders tiles viewable by Security's cameranet, showing everything else in static (like the Station AI's view). Getting EMP'd will also cover the visor's vision in static for a short while.
+
+## Game Design Rationale
+
+### Aesthetic Consistency
+
+The art and lore teams are trying to establish a more consistent tech level (1980s cassette futurism) across the board, and I believe our design should take that into consideration. When it comes to digital interfaces in the game, they should either be attached to a large, immobile mainframe (such as consoles) or relatively simple in function and implementation.
+
+While heads-up-displays *as a technology* definitely existed in the time period we are drawing from, it was mostly limited to aircraft cockpits, with head-mounted HUDs being cumbersome and experimental (hence the design of the panopticon helm). Of course, this is a science fiction setting where artistic liberties are taken with future technology, but Google-Glass-style eyewear does not fit in with the style of tech we envision the crew using.
+
+### Maximizing Roleplay Potential (Avoid QOL slop)
+
+Almost all HUD eyewear is inherently QOL slop.
+
+Security HUDs exist to save a security officer from spending a single second checking someone's ID. This stifles any opportunity for subterfuge that doesn't involve either taking someone else's ID or using an agent ID. If we removed them, anyone with a reason to be deceptive about their job or position has the chance to sneak past inattentive officers, instead of being busted by a superfluous UI element. Security officers can ask to see someone's ID if they're out of range, which is more interesting roleplay than just instantly knowing.
+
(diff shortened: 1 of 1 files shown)


mkanke-real:
I dig it. I really want to see the ripple effects of this, especially if we can add more ways to subtly hide your identity. It gives a lot more options than just going loud as soon as your identity is known which leaves more RP avenues.

vivry:
I think this has very interesting implications in terms of faking privileged access with sec and command. But I wanted to say that my own proposal, #69, kind of needs the medical HUD, or something that acts like it. The PR is supposed to improve med by removing health analyzers and replacing them with a bigger and slower scanner machine. However, med staff will need the ability to quickly triage patients more than ever, since they now won't be able to just mass scan everyone in the lobby. They need to be able to instantly sort the patients by urgency. This can be done with just a nerfed health analyzer that only shows overall damage (which is what I considered doing and might end up doing if this goes ahead), but the medical HUD already does the same thing, and it's already implemented. Regardless of my PR, med will probably head in the direction of being more crunchy and authentic, with proper diagnostic gameplay, so this will need to be a consideration.

pirakaplant:
> However, med staff will need the ability to quickly triage patients more than ever, since they now won't be able to just mass scan everyone in the lobby. They need to be able to instantly sort the patients by urgency.

Surely the inspect menu works for this? You can get the severities of injuries and their types with two clicks. Some damage types are "invisible", but that's the exception, and patients can just explain they're poisoned or whatever in the off chance that it comes up.

masterelzoi:
while i cant comment for med. i think this is a terrible idea for security. This is going to introduce the rev issue for all other conversion antagonists. With no way to verify mind shield status, criminals are going to be getting dragged back to sec repeatedly, unless communication is on point 100% of the time, which lets be honest isn't gonna be the case in most rounds. I would hate for blood and cosmic cult rounds becoming TDM because security cannot trust anyone who isn't shielded, which now extends to everyone, as we cant tell for anyone. I think this is needless removing a feature, when there is already a lot of ways to evade the sechud. At the minimum a device needs to be added in place to verify mind shield status, even if is a big stationary thing in sec. This is also inconsistent with the way borgs function and how they identify crew. This will also encourage power gaming in sec, as you CAN identify mindshields by attempting to inject someone with a mind shield will only allow it if they aren't shielded. I think sec carrying around mindshields to check especially in conversion antags is a bad idea.

DeltaSS14:
> Surely the inspect menu works for this? You can get the severities of injuries and their types with two clicks. Some damage types are "invisible", but that's the exception, and patients can just explain they're poisoned or whatever in the off chance that it comes up.


So, I'm not in favor of removing the medhud alltogether. I am much more in favor of the reworked idea for a medical scanner, but keeping an easy and quick indicator for medical information. Sure the inspect button works right now, but it's rather clunky (why is it not just on the immediate page for inspecting someone) and has various damage types missing (you can't ask em if they are crit or dead, and often times people don't know why they are hurt.) With that, I think you *should* be able to look around a room and get a vague idea if someone is utterly beat to a pulp or has a minor bruise. Inspecting just doesn't allow you to do this quickly, and it's also not visible on sprites enough to do it without some kind of indicator.

I understand that the medhud feels very much "video game health bar", but I'd rather see a proposal to rework them other than just outright removal. Something like removing the health bar but keeping a status indicator for example (stable, crit, dead), or allowing people to put indicators on patients for other docs to see and show priority (RMC has this.)

I am very much in favor of changing the medical system, and something like conditional med would fix a lot of the gamified issues med has, I don't think just outright removing medhud accomplishes much of anything to be honest. At the end of the day, shitmed is a system based on a single health bar, just trying to pretend it isn't won't fix the system.

I think the idea of being able to apply indicators onto patients, that people with a medhud can see without inspecting, would be the best way to go about it. This would make medical actually triage, and force more communication within the department.

Catazeus:
I like this proposal, however, the only issue with it I can see is lack of ability to distinguish mindshielded personnel. As I put in my similar proposal I posted on Funky Discord server, mindshielded people that are flashed will have their eyes light up.

However, to embrace sec redux and avoid powergaming/metagaming, flashes need to cause actual negative effects on eyes. And the lamp in interregation room would need to be beefed up so it isn't taken by every secoff.

**To sum it up:**
- Flashing people with a mindshield or (enabled) fake mindshield will make their eyes flash white with a context popup ("_Urist's eyes flash white_").
- Flashing people excessively (>2 times in short span) causes stacking, curable, permanent bit of blindness (go get oculine).
- Lamp in interrogation room is apparently an infinite flash! Make it 4x4 and large so it doesn't fit easily anywhere.

pirakaplant:
@masterelzoi 

> i think this is a terrible idea for security. This is going to introduce the rev issue for all other conversion antagonists. With no way to verify mind shield status, criminals are going to be getting dragged back to sec repeatedly, unless communication is on point 100% of the time, which lets be honest isn't gonna be the case in most rounds. I would hate for blood and cosmic cult rounds becoming TDM because security cannot trust anyone who isn't shielded, which now extends to everyone, as we cant tell for anyone.

I think the warden or HOS should be doing bookkeeping concerning who gets a mindshield, and that person should be doing the mindshielding or calling the shots on who gets mindshielded. If communication issues arise from that, so be it, that's part of the game.

> I think this is needless removing a feature, when there is already a lot of ways to evade the sechud.

This is the same argument that people used against our greencomms changes, particularly with the radio jammer. You shouldn't need specialised equipment or abilities like an Agent ID to get around the magic cop glasses that don't even fit with the tech level we're going for.

> This is also inconsistent with the way borgs function and how they identify crew.

Borgs can just get a "crew"/"not crew" icon instead of the job icon. Also, whole machine with its own visual processing system being able to read IDs is more authentic than a thin sheet of glass doing so.

> This will also encourage power gaming in sec, as you CAN identify mindshields by attempting to inject someone with a mind shield will only allow it if they aren't shielded. I think sec carrying around mindshields to check especially in conversion antags is a bad idea.

We can make the implant removal backfire trigger if something like this happens.

@DeltaDnDsolutions

> Sure the inspect button works right now, but it's rather clunky (why is it not just on the immediate page for inspecting someone) and has various damage types missing (you can't ask em if they are crit or dead, and often times people don't know why they are hurt.)

I wouldn't be opposed to putting it on the main tab (along with ID) if it doesn't bloat up the description too much.

You can use medtek/analyzer for the latter, and if we end up replacing that with something bigger, toxin and genetic (which I'm pretty sure are the only two damage types that don't have a description) are edge cases enough that there wouldn't be a major inconvenience caused by removing medhuds.

> I understand that the medhud feels very much "video game health bar", but I'd rather see a proposal to rework them other than just outright removal. Something like removing the health bar but keeping a status indicator for example (stable, crit, dead), or allowing people to put indicators on patients for other docs to see and show priority (RMC has this.)

Stable/Crit/Dead is something that pretty much everyone can see with their eyes, and as a result doesn't need a HUD in the first place.

In theory I like the tag idea, but any device that can do this that isn't a big bulky machine is going to feel inauthentic to our art direction. These could be physical tags that actually attach to the patient or their bed instead.

@Catazeus

I actually like your idea thematically, and if security bookkeeping mindshields is actually too much of a hassle this is what I'd go for. Maybe the fake mindshield implant could be replaced with smart glass contacts that fake the effect.

DeltaSS14:
> Stable/Crit/Dead is something that pretty much everyone can see with their eyes, and as a result doesn't need a HUD in the first place.

I disagree with this. Given the amount of people that drag dead patients onto normal beds instead of stasis. I feel like if you as a doctor take a glance into the medical ward, you should be able to tell quickly who is dead, crit and stable. Sure you can shift click everyone, but that is much more tedious in a way that doesn't feel authentic.

> In theory I like the tag idea, but any device that can do this that isn't a big bulky machine is going to feel inauthentic to our art direction. These could be physical tags that actually attach to the patient or their bed instead.

Yeah, if we take the step of removing huds and going by tags then I think physical tags would be the way to go. They should be something you can clearly see on the sprite though, not just an inspect option. (think irl triage slap bands)

pirakaplant:
Sorry, been sidetracked.

@DeltaSS14

> I disagree with this. Given the amount of people that drag dead patients onto normal beds instead of stasis. I feel like if you as a doctor take a glance into the medical ward, you should be able to tell quickly who is dead, crit and stable. Sure you can shift click everyone, but that is much more tedious in a way that doesn't feel authentic.

It's not really tedious at all. It's literally a single input that instantly tells you everything you need to know in that regard. A medical scanner is more tedious than that. If anything, it's more difficult to tell that someone is dead or in a critical condition in many cases IRL (and realistically this is something the stethoscope could be used for if we decided to get rid of the dead inspect markers for whatever reason).

> Yeah, if we take the step of removing huds and going by tags then I think physical tags would be the way to go. They should be something you can clearly see on the sprite though, not just an inspect option. (think irl triage slap bands)

Yeah definitely.

PR #76, opened 2026-03-03
thief redesign concept

stormyatt:
very simple
1. The objective is now to put 10,000ish spaceos worth of stuff inside a bag they start with that has a ton of space
2. Pacifism is replaced with an objective not to kill

all changes are related to these two ideas

### DIFF
--- a/src/design-proposals/thief redeisgn concept.md
+++ b/src/design-proposals/thief redeisgn concept.md
@@ -1,0 +1,64 @@
+\#Thief Rework 
+
+| storm| Implemented | GitHub Links |  
+|---|---|---|  
+| stormyatt | :information\_source: Open PR and :x: No | TBD |
+
+\#\# Overview
+
+It first replaces thiefs objectives with a objective to put 10,000ish spaceos into a bag. It also replaces the pacifism implant with a objective that will redtext if you kill anyone
+
+\#\# Background
+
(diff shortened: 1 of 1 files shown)


LilHoss4k:
hmmmm. Ehhhhhhhhh. 1st half would be pretty cool, 2nd half I don't support.

Also I think you neglected the proper format.

KittyCat432:
Theres currently not a way to track who killed someone as theres so many ways to skirt around this and so many edge cases.

Also a bag completely removes taking bigger objects.
We have a thief beacon still in code that would work better.

stormyatt:
On the no kill objective, I believe that pacifism can also be subverted. I at least suggest trying it out and simply reverting that part if notable disliked.

on the thieving beacon, it was not used simply because I saw the amount of objects one would need to furfill the objective would likely cause the area around the thieving beacon to become so cramped as to push out certain objects. As well as just be easy to notice.

While I prefer the bag from a personal perspective of the thief having the items they wished to steal on them, if you so wish could we not just make the beacon have a component to allow objects to be tossed inside it, similar to a disposals shute.

stormyatt:
Further note: I forgot to say within since it was not about how it was design but rather balancing. If a thief gets cargo tech or, if they’re allowed to have bigger objects a can of frezon, atmos. Their objective becomes incredibly easy. They could easily scim of the top of cargo deliveries as a cargo tech or make a can of frezon if atmos. 

I do not know how one work around this of if you wish to. I just remembered this and wish to give warning.

PR #78, opened 2026-03-08
CC staff are immune under space law

DeltaSS14:
This proposes  that CC vip's (in particular the NTR and Magistrate) become immune under space law.

### DIFF
--- a/src/design-proposals/CC_Immunity.md
+++ b/src/design-proposals/CC_Immunity.md
@@ -1,0 +1,40 @@
+# Company Scrip
+
+| Designers | Implemented | GitHub Links |
+|---|---|---|
+| Delta | :x: No | TBD |
+
+## Overview
+
+Central Command staff, including CC V.I.P. roles, are fully immune under space law.
+
+## Background
+
+Central Command staff are outside the chain of command of the station. Part of their role is to be a direct representation of the overwhelming power that Nanotrasen has over the people on station. In discussions about how security should be dealing with command and CC staff it's been concluded that security should not be arresting these people in almost all cases, seeing as security's primary purpose should be protecting the interests of the company. 
+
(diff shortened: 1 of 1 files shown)


Idem778:
So, one thing, IAAs should definitely be included in this immunity because they are also said to the SoP frontline, if a class one has to be delivered the IAA is send, if sec is to be inspected an IAA is send, the CC VIPs that pretty much always face spacelaw for doing their job and following orders are the IAAs
Oh and captain SoP 8 still exists, so there is one thing to hold CC VIPs accountable

ferynn:
Not a fan of this change. Space law is meant to be Nanotrasen law, in how they run and organize the station.

Members of command/central command shouldn’t be immune to it. That’s not to say it can’t be or shouldn’t be _biased_ towards them, or even unfair. Just that the unfairness should be represented in the laws themselves, rather than an immunity to them.

DeltaSS14:
> the CC VIPs that pretty much always face spacelaw for doing their job and following orders are the IAAs

This is sort of exactly why I chose not to include IAA in here. This change isn't meant to come up often, and it comes with a certain degree of scrutiny that those affected aren't shitters about it. I want IAAs to be the overzealous attack dogs of the NTR, but not constantly brushing up against the law.

> Oh and captain SoP 8 still exists, so there is one thing to hold CC VIPs accountable

Good point, in this change that sop would also have to go. Captain and Security *can* still detain CCvips, it'll just be an overt and direct breach of norms.

DeltaSS14:
> Just that the unfairness should be represented in the laws themselves, rather than an immunity to them.

The way I might word it is that local security isn't to arrest CCvips, but instead report unlawful behavior to Central Command to be investigated. It just doesn't make sense to me that a company would allow local security to arrest their representatives on their own accord, what are your thoughts on that?

PR #79, opened 2026-03-11
Mage of ascension

Terkala:
Adding a design doc for a combination Wizard/Heretic for Forky.

### DIFF
--- a/src/design-proposals/mage-of-ascension.md
+++ b/src/design-proposals/mage-of-ascension.md
@@ -1,0 +1,114 @@
+# Mage of Ascension
+
+| Designers | Implemented | GitHub Links |
+|---|---|---|
+| Terkala | :x: No | |
+
+## Overview
+
+This proposal adds a new Mage of Ascension antagonist built around a Ley Line system and spell schools. Mages of Ascension start weak and grow into a nearly station-ending threat by opening confluences (ley-line confluences) to places of power. Spells are organized into schools (Necromancy, Elementalism, Honkamancy, Hemomancy, Dimensionalism, Chaosmancy), and progression is tied to confluence-opening and a mana system. This variant exists alongside the existing Wizard as a midround/event antagonist.
+
+**Player goals:** The Mage's goal is to ascend to ultimate power. Once ascended, their goal is to show the station their power.
+
+## Background
+
+**Relevant code:** `Resources/Prototypes/Catalog/spellbook_catalog.yml`, `Content.Shared/Magic/SpellbookSystem.cs`, `Resources/Prototypes/Entities/Objects/Magic/books.yml` (WizardsGrimoire), `Content.Server/BloodCult/EntitySystems/BloodCultRiftSetupSystem.cs` (3x3 confluence reference), electrical anomaly spark behavior for mote travel.
+
+## Features to be added
+
+### 1. Ley Line & Confluence System
+
+Confluences are ley-line confluences—points where ley lines converge, opening rifts to places of power.
+
+**Spawn rules:**
+- One confluence exists at round start.
+- One new confluence spawns each time a confluence is pried open.
+- Only one "new" (unopened) confluence exists at a time.
+- Confluences spawn indoors, in areas with safe atmosphere (normal temperature, normal pressure).
+
+**Visibility:**
+- Unopened confluences are visible only to the mage.
+- When a confluence is pried open, its sprite changes and it becomes visible to all players.
+
+**Finding confluences:**
+- Each confluence fires a mote of mana on its own timer (25–35 seconds, randomized to prevent syncing).
+- **With only one open confluence:** The mote shoots from the unopened confluence toward the mage. The mage can trace the vague direction to find it.
+- **With two or more open confluences:** Each confluence randomly shoots its mote toward either the mage or another confluence. The mage must either stay near a confluence to observe which direction the mote travels, or hope to get lucky. Mana motes travel through walls (similar to electrical anomaly sparks).
+
+**Opening confluences:**
+- The mage must have the spellbook out and open in one hand.
+- The mage clicks the confluence with the open spellbook.
+- Once pried open, the confluence's sprite updates and it becomes visible to everyone.
+
+**Final confluence:**
+- Special, more visible confluence.
+- Occupies a 3x3 area (similar to blood cult final rune rift).
+- Grants the capstone spell when opened.
+
+### 2. Spell Schools
+
+Six schools: Necromancy, Elementalism, Honkamancy, Hemomancy, Dimensionalism, Chaosmancy.
+
+**Selection:**
+- The mage chooses a school at round start via the Grimoire.
+- This locks their spell path.
+- After choosing, the path store transfers to the mage (similar to uplink implant).
+- The Grimoire is then used to interact with confluences to drain them.
+
+**Progression:**
+- Each confluence opened: choose between two spells for that school.
+- Spells improve by tier as more confluences are opened.
+- Capstone spell: earned after opening the final confluence.
+
+**School themes (high level):**
+- **Necromancy:** Undead, death, resurrection. Capstone: transform into an immortal lich that reforms their body minutes after death.
+- **Elementalism:** Ice/fire. Capstone: constant ice effect around the mage, replacing floors/walls with ice and freezing nearby players.
+- **Honkamancy:** Clown/silly. Capstone: transform into a Clown Behemoth roughly as strong as a dragon.
+- **Hemomancy:** Blood/vampiric.
+- **Dimensionalism:** Teleportation, confluence manipulation.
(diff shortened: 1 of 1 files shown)


masterelzoi:
So far so good! Do like the ideas for sure. I think we do have to keep the dramatic server wide messages once you unlock your capstone. it is a good feeling for peoples story as the antag to reach a peak, and everyone feel like they can get involved. Can their progression be slowed or stopped by security at all other than arrests? Additionally if you do want to go down the route of people being able to tell vastly different and in depth stories, with the unique identities. I feel a path covering what heratic did is missing. The only one that comes close in my opinion is dimensionlism, even then would need to see more of it. Lichdom does not fit an eldritch feel as much as it does a gothic one. Worth perhaps pushing down the very esoteric incomprehensible route with dimensionalism

Terkala:
> So far so good! Do like the ideas for sure. I think we do have to keep the dramatic server wide messages once you unlock your capstone. it is a good feeling for peoples story as the antag to reach a peak, and everyone feel like they can get involved. Can their progression be slowed or stopped by security at all other than arrests? Additionally if you do want to go down the route of people being able to tell vastly different and in depth stories, with the unique identities. I feel a path covering what heratic did is missing. The only one that comes close in my opinion is dimensionlism, even then would need to see more of it. Lichdom does not fit an eldritch feel as much as it does a gothic one. Worth perhaps pushing down the very esoteric incomprehensible route with dimensionalism

I'd love some suggestions for ways security could slow them down other than an arrest. Realistically, I don't think security will be able to detain them without lethals, due to their ability to escape with their blink, so it'd start from a situation in medbay of them being revived from crit/dead. I could tie it to an item, but I want them to have the ability to blend as crew without having an item giving it away, and don't like the jank of having to equip/unequip an item to teleport.

So my initial thought is simply to have the same high risk/reward structure of Wizard, where if caught you're executed because it's too difficult to contain them. But would welcome any ideas for how to handle it differently.

I was actually thinking of slicing up heretic abilities to use for other paths. Hemomancy/Necromancy would use a lot of the Flesh Path assets. Dimensionalism would use void path assets. Elementalism would use some of the ashen path assets. I can't use the actual spells or code, since it's AGPL, so at best I'm just using sprites and icons. And I want to intentionally avoid using Runes since I want to help streamline mechanics and those end up being confusing for players to utilize.

masterelzoi:
Forced executions when catching a certain type of antag will never be a good idea. All it does is make corporate secrets reveals happen far too frequently as evidenced with changelings so i would definitely avoid that. You could work it in with the oubliette or similar device to either reset or wipe their magic capabilities completely. I think its fine for them to just have the teleport without an item as long as cuffs stopped it. This is an antag which has to deal with antags AND sec intrinsically. No other antag faces that challenge, so throwing them a bone is fine. Same way heretics get a grasp they get a teleport. Acquiring custom gear could just be done through the spell path menu. Such as the paths armour (they for sure should get a kind of armour) if you want to avoid using runes entirely, which i think is fair.

Terkala:
That's a fair point, how about this as an idea:

Locked Lodestone Collar (warden gets one in their locker, and lathe can make it out of steel). Neck slot item, which has a security-lock (need security level door access to unlock). So you can lock it onto a captured mage, and they can't use their powers while wearing it. But if they can find (or make) an incapacitated officer after they're released from prison, they can use the officer's ID card to remove the collar. That way they're not entirely doomed, and have some way back if caught, but are "unlikely enough" to get access that security doesn't feel like they need to find an excuse to perma them

DeltaSS14:
> Locked Lodestone Collar

I like the idea of a lockable item doing this, may want to avoid a collar since people will be weird about it (see shock collar removal)

Terkala:
> > Locked Lodestone Collar
> 
> I like the idea of a lockable item doing this, may want to avoid a collar since people will be weird about it (see shock collar removal)

Lodestone Necklace? I want it to be neck slot so it doesn't block hands, so it's something they can just release the prisoner with but the player will need to find a sec-id to remove

masterelzoi:
> > > Locked Lodestone Collar
> > 
> > 
> > I like the idea of a lockable item doing this, may want to avoid a collar since people will be weird about it (see shock collar removal)
> 
> Lodestone Necklace? I want it to be neck slot so it doesn't block hands, so it's something they can just release the prisoner with but the player will need to find a sec-id to remove

Make it a glove slot item. Gloves that cant be removed, allows hands to be used but takes the glove slot

Terkala:
Like a reverse power glove? Magic Suppression Gloves? Yeah that would work.

BalkanCalm101:
> And I want to intentionally avoid using Runes since I want to help streamline mechanics and those end up being confusing for players to utilize.

You could use Orbs instead of Runes.Fits the wizard theme pretty nicely.
I also hope Mages don't get depowered from the Oubliette since,uh...it just kind of sucks. Antags that have the power to destroy the station should be dealt only with executions,in my opinion. It just kind of a cope out to use the magic box to magically make the antag start from scratch with an even harder time. Oubliette is worse than an execution.

If you want a different way to go about it,then maybe make their powers tied to an organ tha can be remove through surgery(?). Or maybe their power is linked to Orbs that have to be gathered and destroyed,or something like that. Up to you..

I also hope the collar doesn't just take pure steel to make.I feel like if you really wanna go that route,then you should make it expensive. 2 diamonds + 2 artifact fragments + 10 steel might be too much,but i feel like containing mages shouldn't be so cheap and easy.

Also,instead of making them completely powerless when that collar is on,you coulb possibly just make it so they can only do VERY basic magic that costs a lot of mana. Otherwise, you'd have to fight a secoff with lethals on him and it just isn't that easy for everyone. But that's just my thoughts.

I really do like this PR and i hope it gets merged when the time comes. I just hope Necromancy isn't just a reskin of Flesh Path and it gets actual love (which it deserves).

Terkala:
Orbs are famously very mobile though. So if you're holding an orb, why not just have a spell that makes the item instead? It's very easy to make spells that just summon items, or transmutes one item from another item (requiring you have X type of item in hand, for example). Simpler and more intuitive.

The idea with the Lead Gloves (gloves instead of collar now, so people aren't weird about them) is that security can "reasonably assume" the mage is handled, and release them. And the mage has a chance to remove them, if they can find an ID card with security access. So they can try to jump a security officer in maints, or wait until another antag downs an officer or creates a distraction. The idea is to give security a "good enough" solution where it doesn't feel unreasonable for security to let a criminal go after X minutes in the brig.

Which solves some of the issue with why Heretic had the Oubliette interaction (and I do agree that it would have been better narratively if heretics were just executed), since there's no reasonable world where security would arrest a heretic and let them go to continue being a problem, since they can re-acquire their gear and spells fairly trivially.

BombasterDS2:
Read this and i just think that heretic and wizard have different vibes. I like a lot of ideas (confluences and mana system), yet i think that wizard and heretic should still be separate antags.

Heretic is related with something eldritch and evil. There magic often is horror for crew and all players know that heretics NOT friend to you even if you antag. Their sounds and music fits this thematic. Confluences mechanic work fits very well to them, fitings with their ascension race.

Wizard is other hand is silly guy that makes funny magic, not always good, not always evil. They are instant sources of chaos and magic on station, so you can't be sure - can you safely interact with them and not.

I'm worried about how both antags will feel after they being united. Silliness of wizard can easily overlap all eldritch and horror vibes of heretic. If changes in this doc will become true i just hope that school will determine vibe and sounds on antags (Sometimes you just want to play true eldritch horror of darkness and sometime you want play magic princess that turns borg into PANCAKE BORGS)

BombasterDS2:
Also there might also be fatigue system for spells (like in some games): Some spells might have no cooldown, but their manacost increases every time they used. Fatigue decays overtime. 
There is also another implementation of this system: overall fatique - when you use any spell, fatique increases, increasing manacost for all spells and decreasing mana regen.

These system will provide wizards "second chance" if they teleported in really bad position or if they really need to throw three fireball at this sec. group.

Terkala:
I had an idea for a different gameplay loop for mages:

1. Make mages similar to Traitors. You're a Mage Apprentice, on the station to retrieve things for magical research.
2. You have a list of objectives similar to a traitor. Such as Steal X or kill Y.
3. With your steal objectives, you have to take the item and push it into a rift in order to see the "magical resonance". Which deletes the item and opens the rift. Rifts have no purpose other than to be a place to deposit stolen items.
4. With your kill objectives you use your magic book on the body to capture the soul (optional RR, allowing someone to be revived with a mind and as a "soulless husk" would probably be fine lore-wise).
5. Doing either awards you with a new spell. Killing another mage and capturing their soul with your book also awards you with a new spell.
6. Completing all of your objectives spawns a larger rift. Only one mage can ascend at this rift, making them all inclined to fight eachother over it. You can only use the rift if all your objectives are complete, and doing so is very obvious and takes a long time. Doing this successfully gives you your capstone spell that is roughly equivalent to a heretic ascension.

PR #80, opened 2026-03-12
Old Station Redux

hereelabs:
# Old Station Redux

I'm opening this Pull Request now so the nitty gritty of this can be discussed, since I'm sure there can be plenty of ideas and improvements (and criticisms!) for this concept. This PR adds the Old Station Redux design doc, which centers around bring Old Station back as a way for players to focus on something constructive once they're no longer able to affect the round (through round removal.) If they do well, they will be able to return to the round, with a unique character and setup in tow. This was discussed several months back, and taydeo suggested many of the things implemented such as the raffle system and (I believe?) Telescience integration.

### DIFF
--- a/src/design-proposals/oldstation-redux.md
+++ b/src/design-proposals/oldstation-redux.md
@@ -6,7 +6,7 @@
 
 ## Overview
 
-Old Station returns as a Telescience based side game. Ghosts are prompted to join a raffle, and will be sent onto a variant of an old, derelict station, tasked with making contact with the main station, using Telescience. The crew of Old Station will go through many dangerous dungeons, before finally sending the station to close proximity of the main station where the rest of the round is taking place. This side game will allow players to focus their feelings after being removed from the main round towards something constructive, rather than something negative.
+Old Station returns as a Telescience based side game. Ghosts are prompted to join a party, and will be sent onto a variant of an old, derelict station, tasked with making contact with the main station, using Telescience. The crew of Old Station will go through many dangerous dungeons, before finally sending the station to close proximity of the main station where the rest of the round is taking place. This side game will allow players to focus their feelings after being removed from the main round towards something constructive, rather than something negative.
 
 ## Background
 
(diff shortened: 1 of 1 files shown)


pirakaplant:
I think this would be a fun thing to happen occasionally. Building something like this into the consistent, guaranteed roundflow would feel a bit arbitrary though. Coming into contact with a derelict vessel is a classic science fiction premise. Can you imagine it happening every single book in a series, or every single episode in a season? I would either make it a chance to happen as part of the game mode, or bump up the difficulty to make successful contact rare.

> This side game will allow players to focus their feelings after being removed from the main round towards something constructive, rather than something negative.

I do take particular issue with this kind of design. We've had a few people trying to "fix" round removal via ghost roles, and while it's a commendable effort, it's ultimately counterproductive and reinforces an unhealthy relationship with the game. The idea itself is solid, but if you're coming from the mindset of trying to appease players who cannot tolerate round removal, I can tell you from experience that they cannot be appeased.

> When a player is a ghost without a body to return to, they will be prompted to join a raffle for Old Station. After 5 minutes have elapsed, up to 6 crew members will be chosen, out of the ghosts who joined the raffle

This is a bit different from how ghost roles normally work. I'm assuming this is because Oldstation originally did not get that many people joining, but I feel like people were going to want to play this more.

Also, the wording here is a bit confusing. Is it going to wait for six people to be dead before doing the raffle once? Because I can't really think of any other interpretation that would work.

Besides that, I'm all for this.

hereelabs:
> Building something like this into the consistent, guaranteed roundflow would feel a bit arbitrary though...

While it is always accessible, that doesn't mean it always will influence the main round. People can fail. It can take so long, that you never even see it in the main round. The problem with Old Station before was that it **was** ever present in every round, being a few hundred units away from the station. This was a genuine concern, but I do believe the new changes make this less likely. Hopefully. Still might be worth trying to see if making it rarer would be good.
> I do take particular issue with this kind of design...

I can see where the concern comes from here, and I agree. I think I overblew how much I took this specific aspect into account, though. It *can* help with players who feel that way, but the goal is moreso to make the everpresent incoming shuttle roles into an actual challenge, one that builds the character as you go, rather than just getting a free pass onto the station once more.
> Also, the wording here is a bit confusing...

Yeah, I will go back in to clarify a bit better. I would clarify in this comment, but I feel like it'd make more sense when all together.

hereelabs:
I am considering having the final disk just spawn a portal at a random safe spot on the station, instead of teleporting the whole Old Station into the same map as the main station.

Terkala:
The core idea I like, but I see some issues
1. What about people who join roundstart to take ghost roles. This is something that happened a lot with oldstation sleepers. People would join roundstart to get a "salvage tider" ghost role, where they would quickly complete oldstation and just come aboard as tiders with powerful gear.
2. Most players don't get RRed until the 30min-1hour mark. If the puzzle takes an hour to do, they arrive just in time for evac, which isn't enough time to do much.

A possible solution: Have it tied to the event pool, and give it a minimum time of spawning so it doesn't spawn until the ~30 minute mark. And specifically disable this event for Survival rounds (where plenty of ghost roles are present due to the nature of the game mode).

hereelabs:
> What about people who join roundstart to take ghost roles.

I think there really should be more in-game punishments for observing for a round. Personally, I think you shouldn't be able to join Old Station unless you've been alive for at least 25 minutes, as a ghost role or not. 
> Most players don't get RRed until the 30min-1hour mark.

Fair enough. Though, returning to the round isn't the entire purpose of Old Station, it's just the end goal. Even if you don't make it to the station in time, or you make it to the station just as Evac arrives, you'll still have something to take away from the events of the round since this Doc aims to have Old Station be on it's own mini-round, completely seperated from the events on the station.

Terkala:
> Personally, I think you shouldn't be able to join Old Station unless you've been alive for at least 25 minutes, as a ghost role or not.

Technically something the code can support, if you want to add that to the design doc. Make it so the button is only click-able if you had been alive for at least X minutes during the round.

PR #81, opened 2026-03-16
botany rework proposal

Seva-2:
Adds a proposal outlining a rework of botany


Seva-2:
UI mock up of DNA Manipulator
dna ui

PR #83, opened 2026-03-22
Proposed Tabletop Rework

CoulsTech:
This is my first PR,

Please be gentle.

I have a pet peeve with chess specifically, in the "Tabletop System"

Please advise

### DIFF
--- a/src/design-proposals/tabletop-rework.md
+++ b/src/design-proposals/tabletop-rework.md
@@ -1,0 +1,51 @@
+# Tabletop Snapping Rework
+
+| Designers    | Implemented | GitHub Links |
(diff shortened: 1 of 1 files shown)


Scrambledking:
There is a chem you can probably just rip the code from to do this

CoulsTech:
> There is a chem you can probably just rip the code from to do this

I would appreciate it if you could be more specific of the component, or chem, or function that does the "snap to map grid" that I could use?

I appreciate the feedback!

Scrambledking:
Sorry didn't get the notification, it should be tilenol if memory serves, it forces the entity to the grid every second or so

PR #85, opened 2026-03-28
[FORKY] Atmospheric Consolidation

pirakaplant:
This design document is specifically for implementing the removal of Atmospheric Technician on Forky Station, with plans to smooth over the hole it left (so to speak) and rename, remove, or replace all of its exclusive equipment.

Even though @yaraaray (who made the PR to remove Atmos Tech on the original codebase) did not personally write anything in the document, I listed her as a co-author since I had her read over the document and she approved of it without any edits.

Everything else you need to read is in the doc itself.

atmos_techs


Dusty-Plant:
I agree with all of this design doc, but could we keep the atmos tech plushie? It does no harm existing and I doubt players would use it as an excuse to say atmos techs are still a thing. it could even be renamed to "firefighter plushie"! 
Also, we may want to keep the teal markings for atmos as they not only allow for easy understanding of where the bay is (memorizing the map, understanding the limits of the atmos bay and the space it encompasses for purposes of spacing or gas leak prevention, preventing confusion for people entering engineering etc) but also allow for more variety in colours, and I think aesthetics are important. 
Other than those two tidbits this is seems like a great change! Encouraging all engies to learn how atmos functions and to work on gas production etc might increase the amount of interaction between the rest of the crew and exotic gasses we currently still rarely see in play.

pirakaplant:
@Dusty-Plant sorry I saw this comment around the time you posted it, I just didn't get around to replying.

> could we keep the atmos tech plushie? It does no harm existing and I doubt players would use it as an excuse to say atmos techs are still a thing. it could even be renamed to "firefighter plushie"!

Yeah I'm fine with this personally.

> Also, we may want to keep the teal markings for atmos as they not only allow for easy understanding of where the bay is (memorizing the map, understanding the limits of the atmos bay and the space it encompasses for purposes of spacing or gas leak prevention, preventing confusion for people entering engineering etc) but also allow for more variety in colours, and I think aesthetics are important.

I agree and I'll update this to make it more clear what I expect out of atmos remapping. It's generally the separate front desks that irk me more.

PR #88, opened 2026-04-02
Borg Redesign

Scrambledking:
Does not touch the laws themselves.

After a number of conversations in the discord I feel like its about time to put the coagulated ideas from discord to the docs page.

### DIFF
--- a/src/design-proposals/Borg-Redesign.md
+++ b/src/design-proposals/Borg-Redesign.md
@@ -1,0 +1,63 @@
+# Borg Redesign
+
+| Designers | Implemented | GitHub Links |
+|---|---|---|
+| Scrambleking | :x: No | TBD |
+
+## Overview
+
+This document seeks to provide an option for the future borg rework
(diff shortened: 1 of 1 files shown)


masterelzoi:
In the corporate hellscape that we are trying to take funky, a sentient robot that has to do what you say is much darker than one that cant feel the repercussions of it. Reducing the capacity in which a borg can talk just takes away a lot of potential communication beyond job functions, which i don't think is interesting. Also borgs being an easy antag thing isnt really an issue with Borgs, its an issue with the emag being insanely strong. Also it doesn't make as much sense as borgs are very often actual MMI's, which are humanoid brains in chassis, which makes the laws to brain thing much harder and makes them being able to express discontent much more believable. Valid hunting is also not an issue with borgs itself, as one its a rule break, and two, borgs hunting antags is down to their lawset, not functionality. This is also fixed by just nerfing how tanky borgs are

PR #93, opened 2026-04-22
Medical Rework Doc

Toastermeister:
Proposal for a Medical Rework, specifically to damage and chem healing.


mikethememer:
Does this have how nukie agent + antags in general should work with this? Most antags are solo and without effective ways to self treat yourself might be really bad for them to be able to effectively engage in long-term combat

PR #96, opened 2026-05-08
[SOP Workgroup] Logistics & Service Restructuring

pirakaplant:
This is a peripheral document for the SOP Workgroup's overall redesign of the corporate structure, specifically focusing on the Logistics (Cargo/Supply) and Service departments.


src/design-proposals/logistics-service-restructuring.md
```diff
@@ -0,0 +1,156 @@
...
+
+**Boxer:** This job will be recontextualised, renamed the "Fitness Instructor", and made a part of the Service department proper, with a dark green icon, instead of being a "Civilian" job. Changes to the job itself would be minimal besides changes to job titles, loadouts and overhauling some boxing areas to put less of a focus on the ring itself and more on.
+
+**Chaplain:** Will be made part of the Service department proper, with a dark green icon.
+
+**Chef:** The default title will be changed to "Cook", and the "Sous-Chef" and "Line Cook" titles will be removed. This is because "Chef", "Sous-Chef", and "Line Cook" are all specific hierarchical titles within a kitchen that don't really work as job title options. More Cook slots will be added to maps proportional to their ideal playercount, encouraging the organisation of a "kitchen brigade" to coordinate on meals.
+
+**Musician:** Will be made part of the Service department proper, with a dark green icon.
+
+**Reporter:** The Reporter, as an codified job, will no longer exist. Instead, Reporter areas and equipment will be all-access, leaving crew to publish their own news stories as they desire.
```
Teasq: I don't see this resulting in good roleplay honestly, even with the reasoning below. I think we need a better solution to this other than putting it in the hands of the crew.
"Why would they be allowed to report the truth?" is a question that still stands for crew, even without an actual reporter job present. Why is NanoTrasen allowing crew access to station-wide news tools if they want to suppress the truth? At least with the reporter, there is a thinly veiled excuse of "transparency", where the reporter just gets dealt with when it doesn't happen to go NanoTrasen's way. It's easier to monitor a single job's speech than it is to monitor the entire crew.

Beyond that, this change can and absolutely will result in lower quality articles every single round, which is a headache in and of itself.

src/design-proposals/logistics-service-restructuring.md
```diff
@@ -0,0 +1,156 @@
...
+- The Hospitality Director has their own office near the Service area, with the Service communication console (HOP will no longer have theirs) and Service job board (see below).
+- Hospitality Director requires 2 hours of Service Worker, 2 hours of Bartender, 2 hours of Cook (see "Service Job Changes" below), and 8 hours of total Service department playtime to unlock.
+
+Aesthetically, clothing and decor for the Hospitality Director should focus on a "fine dining" or "first-class hotel" aesthetic to indicate their role and superiority to other members of the Service department. For example, their "default" outfit would ideally be a buttoned uniform and pillbox hat, akin to a valet.
+
+### Service Job Changes
+
+There will be multiple changes made to Service jobs themselves. In alphabetical order:
+
+**Boxer:** This job will be recontextualised, renamed the "Fitness Instructor", and made a part of the Service department proper, with a dark green icon, instead of being a "Civilian" job. Changes to the job itself would be minimal besides changes to job titles, loadouts and overhauling some boxing areas to put less of a focus on the ring itself and more on.
```
Teasq: This could be cool. There are various fitness-related vends and items in the game that as of right now aren't used very often in favour of a boxing ring. I think making the boxing ring areas gyms/fitness zones would be much better for roleplay and general station design.

src/design-proposals/logistics-service-restructuring.md
```diff
@@ -0,0 +1,156 @@
...
+
+**Chaplain:** Will be made part of the Service department proper, with a dark green icon.
+
+**Chef:** The default title will be changed to "Cook", and the "Sous-Chef" and "Line Cook" titles will be removed. This is because "Chef", "Sous-Chef", and "Line Cook" are all specific hierarchical titles within a kitchen that don't really work as job title options. More Cook slots will be added to maps proportional to their ideal playercount, encouraging the organisation of a "kitchen brigade" to coordinate on meals.
+
+**Musician:** Will be made part of the Service department proper, with a dark green icon.
+
+**Reporter:** The Reporter, as an codified job, will no longer exist. Instead, Reporter areas and equipment will be all-access, leaving crew to publish their own news stories as they desire.
+
+**Service Worker:** This will be recontextualised into more of a "waiter" job, and the bulk of their work will be communicating orders and serving food to customers of the Service area. Service Workers can still shadow other Service jobs and fill in when needed, but they no longer start with Bar or Kitchen access.
```
Teasq: This is excellent. Service workers as they are right now are just the ultimate service role despite technically being a service intern role. I personally have had a lot of issues with service workers just pushing themselves into whatever service area they wanted to play this round but didn't get the specific role for.

src/design-proposals/logistics-service-restructuring.md
```diff
@@ -0,0 +1,156 @@
...
+
+**Zookeeper:** Will be removed as the position does not contain enough content, roleplay potential, or even player interest for a whole job. The zoos themselves (on the maps that have them) can still exist for players who want to opt into the role of taking care of the animals there.
```
Teasq: I agree. Zookeepers add nothing to a round and are almost always not allowed to keep the actually interesting animals.

src/design-proposals/logistics-service-restructuring.md
```diff
@@ -0,0 +1,156 @@
...
+
+A corporation like Nanotrasen needs to rely on "bread and circuses" to maintain appeal among its workers. After all, worker loyalty is a hot topic in the real-life business world and is typically treated in a similar manner. While NT employees don't exactly have anywhere else to go, the possibility of striking and other retributive and distruptive acts is an issue that Nanotrasen would want to keep in check.
+
+At the moment, Service is a department that is mostly left to its own devices, as the Head of Personnel is busy with their own duties. Giving Service its own department head, whose responsibilites are in line with their department's, is something that Nanotrasen would do.
+
+### Maximising Roleplay Potential
+
+### Adding More Cooks
+
+The kitchen is a great avenue for job-based roleplay that's never really been explored. Ideally, it would allow for an environment where fast-paced coordination punctuates the round, with multiple different cooks working on different parts of the meals to get them under the waiter's cloche as soon as possible. Adding more people to the kitchen is vital for this angle.
```
Teasq: I don't think this will work out as described, honestly. Meals are not difficult to prepare, and there isn't enough people going through the service area in need of meals where it would require fast-paced coordination.

Obviously, making meal preparation more rigorous or redesigning the need for food and water is completely out of the scope of this document, but I also think that adding more cooks isn't valuable for these reasons. Most stations already have 2, and some kitchens already struggle to accommodate 2 cooks. 
I think increasing the number of cooks should happen when there is an actual need for it, otherwise you'll get similar situations botany falls into where there isn't enough space to accommodate the botanist and it therefore makes the role less appealing to play when there is already one onboard or otherwise very annoying.

src/design-proposals/logistics-service-restructuring.md
```diff
@@ -0,0 +1,156 @@
...
+
+#### The Nanotrasen Award for Excellence in Journalism
+
+Removing Reporter from being a codified job fixes a lot of baggage the job had. If Nanotrasen specifically hires a Reporter, why would they be allowed to report the truth? The Reporter had SOP prohibiting them from slandering the station, but something about this interpretation felt a bit flat. It didn't really generate as much conflict as a journalist being on a corrupt space station would.
+
+By putting reporting into the hands of crew, it helps reinforce the class dynamic we're going for. No specific SOP and expectations for reporting means that more crew are going to publish something that corporate would not approve of, and there are going to be more mysterious "suicides by five gunshots to the back of the head" to drive conflict between crew  and Command/Security.
+
+#### Service Events
+
+Service's events provide a prompt for players to get together in ways they otherwise would not, and further cement Service as a department that facilitates social interaction across other departments.
```
Teasq, line 117: This is a neat idea as long as it doesn't happen too regularly. I would say the perfect amount would be about 2 per shift, where one event happens halfway through and the other happens at the end of the shift, assuming nothing goes catastrophically wrong.

src/design-proposals/logistics-service-restructuring.md
Joaco545: > Meals are not difficult to prepare, and there aren't enough people going through the service area in need of meals where it would require fast-paced coordination.

We believe this will change with the union-mandated breaks. Generating moments where whole departments could go to get something to eat, meaning food will be needed in peaks, and not as a more constant thing throughout the shift we have now.

Teasq: > We believe this will change with the union-mandated breaks. Generating moments where whole departments could go to get something to eat, meaning food will be needed in peaks, and not as a more constant thing throughout the shift we have now.

I don't think it will change anything, 2 chefs can prepare piles and piles of food for this in the span of a few minutes. That's why I said I think it should increase when a genuine need for that is noted. As it stands, chefs already barely get any notice thanks to vending machines, and increasing the hands on an already practically non-existent workload will probably just make the kitchen very annoying to work in.
I don't think the idea is without merit, it's just jumping the gun imo. There hasn't been anything outlined that actually increases the chef's workload, or makes the process of making food more involved that would warrant more than 2 chefs.

Joaco545: We have just uploaded the main PR for the SOP workgroup; In there we did the same thing as the bartender has, but for the chef.
Food may not be left unattended/must be thrown if not eaten by the 15 minute mark.

Some mechanical enforcement may be needed (like food getting stale and rotting if left unrefirgerated) for this to make more sense, but an IAA could get on your ass and fining you for not following proper procedure

src/design-proposals/logistics-service-restructuring.md
pirakaplant: Yeah that's kind of a good point in the first paragraph. Later down the line, I want to make news more lower-tech, which would alleviate a lot of the issues as guerilla journalism would be a lot more feasible and harder to suppress.

Tay specifically wanted this change so I put it in even if current game mechanics don't support it yet. I'll talk to them about this and see if it should stay in as something to be done later down the line or be part of a separate document.

src/design-proposals/logistics-service-restructuring.md
pirakaplant: Yeah I was sort of working on the assumption that we would have a better cooking system by the time these changes got enacted (I think one of the other microcosms were working on it). I'll probably either remove it or put a note stating that this would be a prerequisite to make that change.

src/design-proposals/logistics-service-restructuring.md
pirakaplant: Yeah, this is covered in the Roundflow & Player Interaction section.

src/design-proposals/logistics-service-restructuring.md
pirakaplant: Spoke to Tay about it. Removing as it requires changes to the news system currently out of scope, but I'm putting a pin in it.

src/design-proposals/logistics-service-restructuring.md
pirakaplant: Removing this as it requires changes to the cooking system to be feasible.

src/design-proposals/logistics-service-restructuring.md
mkanke-real: I cannot say I agree with reporters removal and would rather it be pushed into being an RP propaganda role than outright dissolution, this is partially as bias as while it is a role I dont play, I enjoy immensely when I interact with it in round. While I very much like the concept of low-tech news, just dissolving it into full crew mush just feels cheap for the various ways the job could be reworked to further the theme that is desired.

src/design-proposals/logistics-service-restructuring.md
mkanke-real: I definitely agree that the change in name should help broaden the category of RP that players will feel confident in taking as I do feel the current name makes them feel pigeon holed

src/design-proposals/logistics-service-restructuring.md
mkanke-real: On a tech side of this I do feel their perms need nerfed a bit to go with this so they do not fall into the same tendencies as the current implementation. Definitely at LEAST botany access removed to keep up with the change of it being switched to logistics.

src/design-proposals/logistics-service-restructuring.md
mkanke-real: While I am sad to see it go its understandable why you want to remove it as its a very bare bones implementation. If I were to change it instead of remove it, the jobs duties would be redesignated as the one in charge of ALL non-sentient life on the stations welfare, this includes checkups on station pets, interaction with botany for animal husbandry and interaction with medical as a form of vet. They would also be in charge of handling and delivering any new animals that cargo buys (such as the recent round with a crab-splosion). This could go so far as to requiring permission from this new role before purchasing new non-sentient life though logistics. 

I believe this is a much more fitting end that requires very little to implement rather than the nuclear option of just flat out removing an already rare job. This is not me saying every station would need to the role, especially not small stations but it leaves room for growth and choice. If you absolutely MUST feel the need to remove it, at the very least make it title options for "Vet" for medical and "Rancher" for botony.

src/design-proposals/logistics-service-restructuring.md
pirakaplant: This has already been removed from the design document for now. As for later down the line, the state of reporter being a codified job does not support the kind of roleplay we ("we" including @taydeo) want from the position, and there isn't really any way to fix that while keeping it its own job.

src/design-proposals/logistics-service-restructuring.md
pirakaplant: I lowkey forgot they had Botany access 🥹

src/design-proposals/logistics-service-restructuring.md
pirakaplant: I genuinely think the amount of people who would miss Zookeeper of all jobs is in the single digits. I believe we can remove it, and if those alternate interpretations of the job are sorely needed, we can just design those from the ground up rather than building them on top of the most forgettable job in the entire game.

mkanke-real: Kinda of rude way to put it but have yours how you like

src/design-proposals/logistics-service-restructuring.md
mkanke-real: I thought the removal from the document meant removal altogether. Is there any written understand of what exactly people including Tay would want from the job as this is kind of a non-answer. Just a brief understand of what is desired instead of a simple no does wonders to actually understand why decisions are being made as right now it just feels like pure vibes with no real reasoning behind it.

pirakaplant: Check the initial commit. I wrote rationale for it that's no longer present in the current document because the context for the reasoning has been removed.

mkanke-real: I mean the initial commit is what I had read first. What I am saying is that omission looks like removal in the context that everything else is still listed. If it is indeed not being removed as a job then there should be a section stating its job since this it to be a comprehensive list of what everyone should be doing.

pirakaplant: I took out the section saying that it would be removed. That section isn't supposed to list every single Service job, only those with changes to be made to them. Given that the removal of another job is still explicitly listed there, omission can't really be read as "this job will be removed".

mkanke-real: I understand, I am just saying what it looks like to someone reading the doc in a vacuum. Im not arguing what you are doing, just giving you clarification on how it could possibly be read without context since every single other service job is listed.

PR #105, opened 2026-05-28
[antag] hitman

stormyatt:
its been a while and a lot have things have changed. take a look if you want


taydeo:
This antagonist doesn't really change up much when compared to a traditional traitor objective. The idea of a more limiting kit is a lot more fun, and something that should be thought over for traditional traitors, but not something stand alone.

I do not see myself accepting this antagonist concept as it is. The entire premise of "specific traitor objective made into antagonist" is not compelling, since it could very much just remain as a specialized traitor objective.

PR #112, opened 2026-06-14
[SOP Workgroup] Paperwork

Scrambledking:
This is a peripheral document for the SOP Workgroup's overall redesign of the corporate structure, specifically focusing on the intended design and implimentation of paperwork.


src/design-proposals/Paperwork.md
```diff
@@ -0,0 +1,102 @@
...
+
+This PR focuses on the purpose and implementation of paperwork within the greater SOP workgroup rework.
+
+## Background
+
+Currently paperwork already plays a reasonably significant role on funky, whether its collecting copies of order forms as QM, IAA/NTR/Magistrate reports, Anomaly report documents, or the significant amount of paperwork that the HoP hands out every shift.
+
+This document may use terms such as
+
+XO Executive Officer (The paperwork half of HoP)
```
pirakaplant: Might just be a nitpick but I'd also mention that XO would also cover accesses like the HOP, and to my understanding of our chain of command, would be the anchor between Internal Affairs (IAA/Magi) and the Captain.

src/design-proposals/Paperwork.md
```diff
@@ -0,0 +1,102 @@
...
+
+#### Fillable Form Fields
+
+Fillable Form Fields are designated areas within "form" documents to be filled in that does not require searching through the often incomprehensible paper formatting code.
+
+This will allow players to fill in Short Answers and Checkboxes for given questions with ease.
+
+#### Command Stamps that flip
+
+Provides a toggle on command stamps for a [Command X] [Command X Approved] and [Command X Denied] stamp variety.
```
pirakaplant: Honestly you'd only need the latter two, in my opinion. Stamps are only really used to mark whether something is approved or denied anyway.

src/design-proposals/Paperwork.md
```diff
@@ -0,0 +1,102 @@
...
+
+### Other
+
+This covers any paperwork that does not fit into other categories. Primarily this will be confronting paperwork that people wish to add, or commonly use, but are not required.
+One such example of this is "request forms", allowing you to formally request something from a department such as a lathe print, or room construction.
+
+### The Mime
+
+The mime is granted leeway around forms due to writing being an oath violation, which would result in their firing. 
+As such the relevant staff or command staff member will have to fill out required forms on their behalf.
```
pirakaplant, line 73: I don't think this would have to be a command thing necessarily. I'd let the mime pick any "interpreter" they desire for these scenarios.

pirakaplant (commented):
This looks really good to me.

As an overall thing for the documentation, I'd like to see some examples of what these forms would look like, maybe even as a "style guide" for how pre-written paperwork should look.

src/design-proposals/Paperwork.md
Scrambledking: While I would agree, I think it is more important to have the RAW specify that the relevant command must do it. That way there is always someone who is available for it.

Additionally its very in theme for the "comedy" role to be able to have a paperwork filled in by another person, only to result in them finishing it, having the mime stamp, then the person who wrote it giving it a **DENIED** stamp.

src/design-proposals/Paperwork.md
Scrambledking: I think there is value in a basic stamp as a sign of "I have read this document" especially for report style forms.

src/design-proposals/Paperwork.md
Scrambledking: Im just deleting this portion as it was left over from initial writing, and does not use the terms.

src/design-proposals/Paperwork.md
pirakaplant: Honestly, I'd call that a form of approval in and of itself. I'm not entirely opposed to it on principle, but I still feel like it's a bit redundant.

pirakaplant: Coming back to this as a maintainer, this is basically the only issue I have with the document. If you either just make it an approved/denied toggle with no ambiguous in-between state or can justify there being such a state, I'll merge this.

As I said earlier, I don't really think there's a scenario where it would make sense but an approved stamp wouldn't. Like with a report, you would use XYZ JOB — APPROVED to indicate that it's been read by you and it's up to your standards, or XYZ JOB — DENIED if for some reason it isn't.

Scrambledking: Examples:

For RD, when they receive a finished Anomaly/Artifact document they do not "approve" the document because there is nothing to approve, they "recognize" they have seen the document.

For CMO, when they receive a Printed scanner report they do not "approve" the damage report, they recognize that they have seen it.

For HoS, they do not "approve" an incident report, they recognize that they have seen it.

For QM, they do not "approve" of a completed delivery slip, they recognize that it has been received.

For CE, they do not "approve" of a repair report, they recognize that they have seen it.

While signatures work in a similar fashion, in the case of promotion, demotion, cryoing, ect. it matters not who recognized it, but that they were in a command position at the time.

pirakaplant: Sorry for the late reply. I discussed this a bit more with the maintainer team and we decided that alternate-mode stamps aren't exactly what we'd want. Command members should instead have a "stamp box" with both their standard stamp and a stock "denied" stamp. If the document could be updated accordingly, that would be great.

Scrambledking: Updated acordingly

pirakaplant (approved):
Image

PR #113, opened 2026-06-14
[SOP Workgroup] Command

Scrambledking:
This is a peripheral document for the SOP Workgroup's overall redesign of the corporate structure, specifically focusing on the intended design and implementation of command and their items.

This does not include the HD.

This does include the XO, and the CL.


src/design-proposals/Command_Design.md
```diff
@@ -0,0 +1,206 @@
...
+This provides the CMO with a tool to assist in emergency situations and mistakes without encouraging them to take over medical entirely.
+Additionally this still makes it a potent tool for antagonists to obtain.
+
+### New Tools
+
+Room Tags : An nameplate/displayboard system to designate rooms by number, doctor, ailment, or more.
+
+### Defensive Measures
+
+Hypo (Space Bear Spray) : Alternate mode for the Hypo, also consuming charges.
```
pirakaplant: The theming on this feels a bit off to me. I like the ability but the name conjures up the mental image of the CMO fighting bears and I don't think that really makes sense.

If you gave it a more generic, chemically name (spitballing and mashing together "omachirine" from Ancient Greek words as an example), I think that would work better.

src/design-proposals/Command_Design.md
```diff
@@ -0,0 +1,206 @@
...
+The spray is similar to a spray bottle, however the chemicals slow the target on contact for a short amount of time, dealing minor caustic damage and blinding them for a second.
+Given the loss of the Hypo and Syringe Gun as defensive options, the CMO will require a new minor defensive tool which this fulfils.
+
+</details>
+
+<details>
+
+<summary> RD </summary>
+
+RD is in a tricky position,
```
pirakaplant: Not sure if this was still a draft that was accidentally marked open or if your document didn't save properly or whatever, but you should probably finish this sentence.

src/design-proposals/Command_Design.md
```diff
@@ -0,0 +1,206 @@
...
+
+Handheld Power Monitor : Altered for additional functionality
+
+### Altered Tools
+
+Handheld Power Monitor (Handheld Engineering Monitor) : Additional tabs added for atmos, nuclear reactor, and supermatter monitoring.
+
+### New Tools
+
+Construction Projector : Shares placed blueprints as Holograms.
```
pirakaplant: I assume you mean by "blueprints" they can display where stuff like walls, windows, etc. should go? You should probably elaborate a bit.

This is more of an aesthetic issue on my part, but I don't know how to feel about devices that project free-standing holograms into the world. Stationary holograms that require constant projection from a device are fine, but something like this kind of pushes belief.

I think an alternative might be some sort of drafting chalk/pen (like a neon one that glows cateye style) that's used to mark out where stuff is supposed to go, with symbols that clearly correlate to specific kinds of wall/window/airlock/whatever sections. Engineers would be able to shift click them to interpret what they mean in the description, so it would be pretty easy to grasp.

src/design-proposals/Command_Design.md
```diff
@@ -0,0 +1,206 @@
...
+Marks items for demolition.
+Allows the CE to effectively direct construction projects.
+
+Remote Cell Charger : Allows the CE to slowly recharge cells in the area.
```
pirakaplant, line 121: Honestly feels like kind of an odd pick for engineering. Science feels like the department that chews up the most power cells to me, and all other departments use it in equal measure. I have no personal problems with it besides that, just felt worth noting.

src/design-proposals/Command_Design.md
```diff
@@ -0,0 +1,206 @@
...
+<summary> Captain/CL (NTR) </summary>
+
+Extra Note:
+In order to better support the Captain being in charge of the station and the Corporate Liaison (CL, previously NTR) just visiting for inspection we plan on the following changes.
+First is that the CL will have their own shuttle that they came/come in on instead of their own room/rooms on the station.
+Second is having the CL come in later in the shift, possibly as a guaranteed event role, to better represent in game how they do not always exist.
+
+### Current Tools
+
+These are appropriately within the theme given their position.
```
pirakaplant: It's probably worth listing these tools, given that they don't currently exist on Forky and would need to be re-implemented.

For my thoughts on them, the cane sword is fine, they don't need the bluespace implant. The handheld fax is a bit unnecessary in my opinion, as they're probably going to be hanging around the bridge when they do their paperwork anyway (and the XO is the one that's going to be getting IAA reports), but there's no game design issues with it.

src/design-proposals/Command_Design.md
```diff
@@ -0,0 +1,206 @@
...
+Satisfied by knuckle dusters
+
+</details>
+
+<details>
+
+<summary> XO </summary>
+
+Extra note:
+The Executive Officer (XO) is the second in command/paperwork side of HoP, split to a separate role with oversight of the Legal minor department involving the altered Magistrate and IAA positions, along with lawyers.
```
pirakaplant: 1. Correct the name of the department to Internal Affairs. Not sure what you mean by it being a "minor" department.
2. Lawyers are not part of Internal Affairs. (They're a Civilian job that mostly works for themselves, they just have a contract with NT that permits them to stay on station to petition clients.)

src/design-proposals/Command_Design.md
```diff
@@ -0,0 +1,206 @@
...
+They generally have a strong connection to the contract-law the station runs on, and accordingly support the station as HoP had previously done before.
+In line with their paperwork and Contract/Legal themes, they still retain the standard duties you would normaly recognise as being within the old HoP's perview, such as setting access and firing/hiring individuals.
+
+They additionally have technical jurisdiction over the Assistants on the station, though they are not part of the Legal department.
```
pirakaplant: Not entirely sure what "technical jurisdiction" means here. Civilians kind of follow their own rules when it comes to who they answer to (Assistants answer directly to absolutely everyone, Reporters answer to their audience).

src/design-proposals/Command_Design.md
```diff
@@ -0,0 +1,206 @@
...
+
+## Administrative & Server Rule Impact (if applicable)
+
+The overall administrative impact should not be overly changed from existing systems.
+
+Rule wise considerations will have to be made for command properly commanding and teaching new players, but these rules do already exist to an extent and so any changes will likely be minor.
+
+# Technical Considerations
+
+UI elements will need to be modified for the various hand held console modifications.
```
pirakaplant, line 218: Some of the features proposed would need more than just UI modifications:

- The portal device changes would require rewriting the C# code that handles the portals
- The change to anomalies would also require C# code
- The RD's handheld research console would require entirely new code for the point tracking thing.

The rest of the features would probably be relatively easy to implement, but those three would have to be the biggest offenders in terms of developer requirements, purely off of vibes.

src/design-proposals/Command_Design.md
```diff
@@ -0,0 +1,206 @@
...
+
+This should not noticeably impact server performance.
```
pirakaplant: The CE's console might. Only one person can look at an engineering console at once for a very good reason: performance. Adding all those bonus functions to the console means that they'll have to be loaded every time the CE opens their console. Like all of them simultaneously.

src/design-proposals/Command_Design.md
```diff
@@ -0,0 +1,206 @@
...
+
+CE Boots : No change, getting to the scene faster to direct engineers is within theme
+
+RCD/RPD : No change, Removed on RCD/RPD overall removal
+
+Handheld Power Monitor : Altered for additional functionality
+
+### Altered Tools
+
+Handheld Power Monitor (Handheld Engineering Monitor) : Additional tabs added for atmos, nuclear reactor, and supermatter monitoring.
```
pirakaplant, line 112: The nuclear reactor and supermatter consoles are both connected to a specific entity in the game world, rather than the power grid or atmos network as a whole. What happens when the CE has the tablet on a station without those power sources? What happens if engineering buys another nuclear reactor?

pirakaplant (changes requested):
These are the issues I have with the document. When they're addressed, I'll approve it.

src/design-proposals/Command_Design.md
Scrambledking: Yea I updated it to "Spraypaint" in the construction doc and never got around to editing it here.

src/design-proposals/Command_Design.md
Scrambledking: At the time this was written it was in a fuzzy position based off the the talk at the time, ill update it acordingly

src/design-proposals/Command_Design.md
Scrambledking: At the time based on what was talked about in the SOP workgroup this was the case, it has since changed.

src/design-proposals/Command_Design.md
Scrambledking: Yep, again, its been over a month since this doc was made and things have settled more, editing it accordingly now.

src/design-proposals/Command_Design.md
Scrambledking: I see, I had assumed it was implemented server side and was single use for the same reason as the crime console, where it was intended to copy a real computer (at least in some sense) since you can technically still open it by spam clicking.

src/design-proposals/Command_Design.md
Scrambledking: _lizardscream.wav_

src/design-proposals/Command_Design.md
Scrambledking: This is different than what I had been told initially about a secondary entity for the SM, which I had assumed would be able to be checked for, upon mapping initialization (or saved for reference by the console) though this in and of itself may have been changing with the SM rework.

As for the reactor, to my understanding it should be able to be linked via multitool.

Either way while I believe the generator links would be nice I do not believe they are required for this doc to function if it is unreasonable.

src/design-proposals/Command_Design.md
Scrambledking: Theoretically in long-term construction engineering should be using work lights, power tools should use cells, and other such uses for cells which this is intended to support.

src/design-proposals/Command_Design.md
Scrambledking: Fair, it was more meant to separate it from the security mace with a real-world comparison.

src/design-proposals/Command_Design.md
Scrambledking: Updating acordingly

src/design-proposals/Command_Design.md
pirakaplant: Yeah, that makes a lot of sense.

pirakaplant (approved):
Yeah this looks good to me now. I'll merge.

PR #128, opened 2026-08-01
Split Personality

vectorassembly:
yum dot com.
i have some bits of the base ready kinda, just felt like I need to ask the others first.

### DIFF
--- a/src/design-proposals/ghost-roles/dissociative-identity-disorder.md
+++ b/src/design-proposals/ghost-roles/dissociative-identity-disorder.md
@@ -1,5 +1,5 @@
 
-# Dissociative identity disorder
+# Split Personality
 | Designers | Implemented | GitHub Links |
 |---|---|---|
 | vectorassembly | :warning: Partially | TBD |
--- a/src/design-proposals/ghost-roles/dissociative-identity-disorder.md
+++ b/src/design-proposals/ghost-roles/dissociative-identity-disorder.md
@@ -29,7 +29,7 @@ Whenever the ghost role is taken, original player (Personal 1) will receive a no
 
 To avoid confusion, some flavor text could be added to a character who is under P2 control.
 
-P1 and P2 can also communicate between each other at all times their own "Headspace" chat channel which cannot be seen by anyone else.
+P1 and P2 can also communicate between each other at all times in their own "Headspace" chat channel which cannot be seen by anyone else.
 
 ## Administrative & Server Rule Impact (if applicable)
 As obvious as it is, this feature introduces a grief vector (like any other ghostrole), to make things smoother, players are encouraged to make some sort of a brief character explanation (which may be added into the character customization menu) to let P2 know how they should (or at least try to) play them, slight deviation is welcome. 
--- a/src/design-proposals/ghost-roles/dissociative-identity-disorder.md
+++ b/src/design-proposals/ghost-roles/dissociative-identity-disorder.md
@@ -1,0 +1,39 @@
+
+# Dissociative identity disorder
+| Designers | Implemented | GitHub Links |
+|---|---|---|
+| vectorassembly | :warning: Partially | TBD |
+
+
+## Overview
+
+This adds a new ghostrole which is something that can be described as Dissociative identity disorder (DID).
+
+## Background
+
+None
+
+## Features to be added
+
+This PR will add a new character quirk which can be toggled on by default, the said quirk can be also automatically added to a character who has suffered a severe head trauma.
+Players can also write a brief explanation of their character's behavior as a hint for the ghost player.
+
+## Game Design Rationale
+
+This ghost role exists on some Space Station 13 servers and can lead to some fun roleplay scenarios, loosing control over yourself and doing things you don't want, etc.
+
+## Roundflow & Player interaction
+
+DID can be toggled roundstart as a character quirk, as well as be a result of a head trauma.
+Whenever the ghost role is taken, original player (Personal 1) will receive a notification, after this, the ghost (Persona 2) can take control over Persona 1 at will (after a certain cooldown of course), upon doing so, Personality 1 can regain control back after a cooldown or let their second personality do whatever they want until the click the specific action.
+
+To avoid confusion, some flavor text could be added to a character who is under P2 control.
+
+P1 and P2 can also communicate between each other at all times their own "Headspace" chat channel which cannot be seen by anyone else.
+
+## Administrative & Server Rule Impact (if applicable)
+As obvious as it is, this feature introduces a grief vector (like any other ghostrole), to make things smoother, players are encouraged to make some sort of a brief character explanation (which may be added into the character customization menu) to let P2 know how they should (or at least try to) play them, slight deviation is welcome. 
+
+# Technical Considerations
+This feature should not affect performance.
+A new text field in the customization menu which explains character behavior might be needed, (could be a multiline text edit in the antag tab?)
--- a/src/design-proposals/ghost-roles/dissociative-identity-disorder.md
+++ b/src/design-proposals/ghost-roles/dissociative-identity-disorder.md
@@ -15,7 +15,7 @@ None
 
 ## Features to be added
 
-This PR will add a new character quirk which can be toggled on by default, the said quirk can be also automatically added to a character who has suffered a severe head trauma.
+This PR will add a new character quirk which can be toggled on, the said quirk can be also automatically added to a character who has suffered a severe head trauma.
 Players can also write a brief explanation of their character's behavior as a hint for the ghost player.
 
 ## Game Design Rationale
--- a/src/design-proposals/ghost-roles/dissociative-identity-disorder.md
+++ b/src/design-proposals/ghost-roles/dissociative-identity-disorder.md
@@ -7,7 +7,7 @@
 
 ## Overview
 
-This adds a new ghostrole which is something that can be described as Dissociative identity disorder (DID).
+This adds a new ghostrole which allows a second person to take over someone's body ocasionally.
 
 ## Background
 


src/design-proposals/ghost-roles/dissociative-identity-disorder.md
```diff
@@ -0,0 +1,39 @@
...
+| vectorassembly | :warning: Partially | TBD |
+
+
+## Overview
+
+This adds a new ghostrole which is something that can be described as Dissociative identity disorder (DID).
+
+## Background
+
+None
```
pirakaplant, line 14: I think you can just omit this entire section if there isn't a relevant background.

src/design-proposals/ghost-roles/dissociative-identity-disorder.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Features to be added
+
+This PR will add a new character quirk which can be toggled on, the said quirk can be also automatically added to a character who has suffered a severe head trauma.
+Players can also write a brief explanation of their character's behavior as a hint for the ghost player.
```
pirakaplant, line 19: I think how this would be implemented would need to be elaborated on. Also, the "can" implies that this is optional. Is this a good idea? There are upsides and downsides to consider.

src/design-proposals/ghost-roles/dissociative-identity-disorder.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Game Design Rationale
+
+This ghost role exists on some Space Station 13 servers and can lead to some fun roleplay scenarios, loosing control over yourself and doing things you don't want, etc.
```
pirakaplant, line 23: I think you would do this idea better justice by expanding upon why we want this for Funky specifically, and split this all up among our game design pillars.

src/design-proposals/ghost-roles/dissociative-identity-disorder.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Roundflow & Player interaction
+
+DID can be toggled roundstart as a character quirk, as well as be a result of a head trauma.
+Whenever the ghost role is taken, original player (Personal 1) will receive a notification, after this, the ghost (Persona 2) can take control over Persona 1 at will (after a certain cooldown of course), upon doing so, Personality 1 can regain control back after a cooldown or let their second personality do whatever they want until the click the specific action.
```
pirakaplant, line 28: Honestly, I think it would be more interesting, purely from a gameplay standpoint, _not_ to alert Persona 1. It would make the alter more of a surprise. Not sure how accurate that would be though.

src/design-proposals/ghost-roles/dissociative-identity-disorder.md
```diff
@@ -0,0 +1,39 @@
...
+
+To avoid confusion, some flavor text could be added to a character who is under P2 control.
```
pirakaplant, line 30: Like, in the inspect popup? I feel like it shouldn't be something obvious to other characters.

src/design-proposals/ghost-roles/dissociative-identity-disorder.md
```diff
@@ -0,0 +1,39 @@
...
+
+P1 and P2 can also communicate between each other at all times their own "Headspace" chat channel which cannot be seen by anyone else.
+
+## Administrative & Server Rule Impact (if applicable)
+As obvious as it is, this feature introduces a grief vector (like any other ghostrole), to make things smoother, players are encouraged to make some sort of a brief character explanation (which may be added into the character customization menu) to let P2 know how they should (or at least try to) play them, slight deviation is welcome. 
+
+# Technical Considerations
+This feature should not affect performance.
+A new text field in the customization menu which explains character behavior might be needed, (could be a multiline text edit in the antag tab?)
```
pirakaplant, line 39: I think there should be a visual mockup that shows where this text field should be and how it should look. Just take a screenshot and draw it in or whatever.

pirakaplant (changes requested):
As the maintainer in charge of traits, this is my jurisdiction.

On the one hand, I think this idea has a lot of potential and I think something like this would be a great addition to the game. On the other hand, we kind of have to address the elephant in the room here:

Dissociative identity disorder is _one of the most maligned mental conditions ever_. For a long time, the only mainstream media representation of DID represented people with it as unstable, dangerous, or even demonically possessed. Even today, those tropes and preconceptions colour how people think about DID and adjacent concepts.

You've managed to avoid the stereotypes that a lot of other depictions employ, which is definitely good, but there is the question of whether the condition can be represented fairly in a gamified form or not.

There are some members of this community who have DID or are DID-adjacent (such as plural systems, a term that includes DID but also encompasses some related stuff to my knowledge). I would ask for their feedback and guidance on this.

src/design-proposals/ghost-roles/dissociative-identity-disorder.md
vectorassembly: It could show a popup for everyone along with flavor text in the inspect thing.

src/design-proposals/ghost-roles/dissociative-identity-disorder.md
vectorassembly: I just think it's fun to willingly allow this roundstart, like, if your character is weird/quirky, or suffers a metnal condition.

you can still get this after head injury whether this is checked or not.

src/design-proposals/ghost-roles/dissociative-identity-disorder.md
vectorassembly: I just feel like people may confuse this for actual player behavior, considering the griefing vector it may be used to make people look bad.

src/design-proposals/ghost-roles/dissociative-identity-disorder.md
pirakaplant: Oh, I meant the brief explanation thing. Like, on the one hand, necessitating it would mean a lot more people playing alternate personas in line with the original personality, but on the other hand, necessitating it puts more of a burden on the player every time they want to make a new character.

Not to mention, not everyone goes in with a personality fully formed! A lot of my characters didn't have a well rounded personality for a bit until I played with them enough to get a feel for how they should act.

So really the choice is up to you there, but it's definitely something that requires careful consideration.

src/design-proposals/ghost-roles/dissociative-identity-disorder.md
pirakaplant: SS14 as a game rests on the fact that people will be doing things for reasons that aren't quite apparent. Antagonists are the classic example, but even emergent conflict can come off as senseless from a bystander's perspective. If someone is griefing using an alternate personality, all there would need to be is some sort of admin indicator or log that shows who's in control, and any ahelp concerning that player can be quickly resolved.

src/design-proposals/ghost-roles/dissociative-identity-disorder.md
pirakaplant: I'm not really sold on that, but could you describe that those would say?

src/design-proposals/ghost-roles/dissociative-identity-disorder.md
vectorassembly: I get what you mean, and while I agree, I feel like brain damage is such a rare ocurance in the game atm, any brain related stuff really, maybe using it as a quirk will do this not roundstart but at some point in round with a chance.

src/design-proposals/ghost-roles/dissociative-identity-disorder.md
vectorassembly: ye true, if we don't think about griefers, this might feel immersive and cool.

src/design-proposals/ghost-roles/dissociative-identity-disorder.md
vectorassembly: not sure myself :catthink:, but maybe something about their movement is being weird, or their eyes being pale.

pirakaplant: tbh none of that really has anything to do with the trait. People with split personalities don't move around weird or have pale eyes.

src/design-proposals/ghost-roles/dissociative-identity-disorder.md
```diff
@@ -0,0 +1,39 @@
+
+# Dissociative identity disorder
+| Designers | Implemented | GitHub Links |
+|---|---|---|
+| vectorassembly | :warning: Partially | TBD |
+
+
+## Overview
+
+This adds a new ghostrole which is something that can be described as Dissociative identity disorder (DID).
```
Gorsthir: While I think this would be fun to add, I do wish it would be called something else, as to not unintentionally reinforce people's misconceptions and stigma surrounding DID.

src/design-proposals/ghost-roles/dissociative-identity-disorder.md
Gorsthir: I do not think there is a need to add anything, as it would just be a matter of how the person acts. Let it just become apparent through roleplay instead.

vectorassembly (commented):
we will call it split personality from now on.

src/design-proposals/ghost-roles/dissociative-identity-disorder.md
vectorassembly: seems fine if we ensure people won't grief, we could add playtime reqs and a proper message.

src/design-proposals/ghost-roles/dissociative-identity-disorder.md
```diff
@@ -0,0 +1,39 @@
...
+
+This adds a new ghostrole which allows a second person to take over someone's body ocasionally.
+
+## Background
+
+None
+
+## Features to be added
+
+This PR will add a new character quirk which can be toggled on, the said quirk can be also automatically added to a character who has suffered a severe head trauma.
```
Teasq, line 18: I think this needs some kind of treatment or temporality to it. It being a permanent condition is a little much, considering that there are many cases where you can receive full-body trauma - which includes the head.

Additionally, while we do have a content warning for loss of control, I think this may be a sore spot for many players and I don't think we can justify a full-round split personality when it isn't opt-in as a condition.

So to conclude, I'd like to see a treatment method present in this section.

Teasq (changes requested):
I just have one pretty major concern when it comes to the head trauma part of this condition. 

Other than that though, I really like this! I can definitely see this system being used for many other things also.

src/design-proposals/ghost-roles/dissociative-identity-disorder.md
vectorassembly: maybe psico pills can suppress or cure it in higher doses? they are kind of a rp fluff as on now.

sleepyyapril (commented):
not a Requirement unless others agree with me but i imagine people would play this up in a sort of "evil alter ego" way and i really would rather that not happen with This specific issue (i don't trust players) so why not just go the "possession target" route and make anything slightly like that not controversial in the slightest?

PR #133, opened 2026-08-23
Energy Weapon & Armory Overhaul

Teasq:
A revision of the lasersec document to bring it up to date with Funky's new design philosophies. Specializes each tool in security's arsenal to promote team play, and specifically gears security to be capable of dealing with crowds through defensive armaments.


pirakaplant (approved):
I'm not going to nitpick every detail given this is a draft, but I already have a few criticisms of the doc so far.

Also, I generally still do not see the point of arming Security with more lasers. There are several things that inherently feel more visceral about ballistics that lasers can't really replicate without being internally inconsistent and not feeling like lasers anymore (e.g. the relative "realism" of the scenario, blood loss, etc.). And if we're making Security less visceral in how they enforce the status quo, that goes against our themes. On top of that, there are balance concerns in reducing Security's damage output to a few specific types rather than a choice of brute or heat damage.

I don't think there are _no_ place for lasers in the game, but it should be balanced out with ballistics in my opinion.

pirakaplant (commented):
Shit, I mean to just comment. I do not approve of this doc yet.

pirakaplant (approved):
Yeah why not.

Radio-Delta:
Related to my previous comments on the x-ray gun, after a conversation on discord yesterday, I don't think the x-ray lance has thematic issues anymore. Not like you need to know I don't disapprove anymore, but yes I'm fine with it and think it's a decent concept.

While I don't think this needs to be included in the design doc, I would advise getting it a resprite and re-sound as soon as is convenient to make it feel less like a generic scifi pew pew laser gun and more like some kind of boxy radiation emitter. Should help with authenticity. The current rendition of it feels a bit like a toy, though the same could be said for every laser weapon in the game, given their sound effects.

PR #135, opened 2026-08-26
Crew: Improvisation, Innovation & Fortune

Teasq:
This is part 2 of a 3 part document which seeks to provide overarching changes to the weapons of Funky Station, to bring the areas touched more in line with antagonist-optional design.

This part focuses on crew weaponry. That which can be improvised, made or found.

### DIFF
--- a/src/design-proposals/crew-weapons.md
+++ b/src/design-proposals/crew-weapons.md
@@ -1,0 +1,116 @@
+# Crew: Improvisation, Innovation & Fortune
+
+| Designers   | Implemented | GitHub Links |
+|-------------|---|---|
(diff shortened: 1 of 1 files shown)


ReconPangolin:
I agree with the general sentiment although the document doesn’t really touch on job specific weapons which I find to be more flavourful and adds some distinction between how different departments engage in combat. 

If part 1 is sec/command and part 2 is crew I’m assuming part 3 would be antag so this would be the document to explore non-standard crew weapons as a part of combat as a whole. Engineering can condense gas to make certain weapons, logistics can buy guns but also grow dangerous produce, medical can manufacture explosives and toxins etc.

